PR #16, opened 2018-11-20
Fetch data from confs-tech and add it to db with command

### LINKED ISSUE #5: Add more data sources
## Todo:

- [x] Create FetcherInterface
- [x] Implement this interface to create a fetcher for https://raw.githubusercontent.com/tech-conferences/conference-data/master/conferences/2019/php.json
- [x] Modify the command `starfleet-conferences-fetch` to let it call the fetcher
- [x] Compute a hash from conference.slug (cleaned: remove year for instance)+conference.startAt+conference.endAt to check if we already have the conference in database
- [x] The fetcher `fetch` method must return an array of Conference entities
- [x] Save conferences in database

We will improve fetchers with other sources in near future.

---
## Help

See:

- https://github.com/tech-conferences/conference-data
- http://api.joind.in/

alexzerah:
Fix #5  

- Add fetcher to confs-tech json data for php
- Add console command action to generate data in db
- Create hash to verify if conference already exists


src/Fetcher/ConfTechFetcher.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+
+namespace App\Fetcher;
+
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\GuzzleException;
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\Psr7\Response;
```
welcoMattic: You call classes from Guzzle, but it does not seem installed. May be you should replace `php-http/curl-client` by https://github.com/php-http/guzzle6-adapter

composer.json
```diff
@@ -16,7 +16,7 @@
         "knpuniversity/oauth2-client-bundle": "^1.15",
         "league/oauth2-google": "^2.0",
         "nyholm/psr7": "^0.2.3",
-        "php-http/curl-client": "^1.7",
+        "php-http/guzzle6-adapter": "1.1.1",
```
welcoMattic: It's better to use `^1.1` notation to let composer update the package to minor version

src/Fetcher/ConfTechFetcher.php
```diff
@@ -0,0 +1,54 @@
...
+namespace App\Fetcher;
+
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\GuzzleException;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Log\LoggerInterface;
+
+class ConfTechFetcher implements FetcherInterface
+{
+    const SOURCE_CONFS_TECH = 'conf-tech';
```
welcoMattic:
```suggestion
    const SOURCE = 'conf-tech';
```

composer.json
```diff
@@ -11,11 +11,12 @@
     ],
     "require": {
         "php": "^7.2",
+        "ext-json": "*",
         "easycorp/easyadmin-bundle": "^1.17",
         "knpuniversity/oauth2-client-bundle": "^1.15",
         "league/oauth2-google": "^2.0",
         "nyholm/psr7": "^0.2.3",
-        "php-http/curl-client": "^1.7",
+        "php-http/guzzle6-adapter": "1.1",
```
welcoMattic:
```suggestion
        "php-http/guzzle6-adapter": "^1.1",
```

src/Entity/Conference.php
```diff
@@ -137,6 +141,16 @@ public function getRemoteId(): ?string
         return $this->remoteId;
     }
 
+    public function getHash()
+    {
+        return $this->hash;
+    }
+
+    public function setHash($hash): void
```
welcoMattic:
```suggestion
    public function setHash($hash): ?string
```

src/Entity/Conference.php
```diff
@@ -137,6 +141,16 @@ public function getRemoteId(): ?string
         return $this->remoteId;
     }
 
+    public function getHash()
```
welcoMattic:
```suggestion
    public function getHash(): ?string
```

src/Command/FetchConferencesCommand.php
```diff
@@ -24,96 +23,33 @@
...
-                    }
-                }
-            }
-
-            if (isset($fetchedConference->cfp)) {
-                $conference->setCfpUrl($fetchedConference->cfp->siteUrl);
-                $conference->setCfpEndAt(\DateTime::createFromFormat('Y-m-d', $fetchedConference->cfp->end));
-            }
-        }
+        $response = $this->fetcher->fetch();
```
welcoMattic: The fetcher returns an array of Conferences now

src/Fetcher/ConfTechFetcher.php
```diff
@@ -0,0 +1,167 @@
...
+
+        $conferencesFinal = [];
+
+        foreach ($params as $date => $technologies) {
+            foreach ($technologies as $technologie) {
+                $client = new Client();
+                try {
+                    $response = $client->request('GET', $this->getUrl(['date' => $date, 'tag' => $technologie]));
+                } catch (GuzzleException $e) {
+                    $this->logger->error($e->getMessage());
```
welcoMattic: You should throw an logical exception here, to avoid next lines be executed

src/Fetcher/ConfTechFetcher.php
```diff
@@ -0,0 +1,167 @@
...
+                }
+
+                $fetchedConferences = json_decode($response->getBody());
+
+                $newConferencesCount = 0;
+                $source = self::SOURCE;
+
+                $conferenceTotal = [];
+
+                foreach ($fetchedConferences as $fetchedConference) {
```
welcoMattic: All this treatment should be moved in another function. To be unit testable in the future

src/Fetcher/FetcherInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * For the full copyright and license information,
+ * please view the LICENSE file that was distributed with this source code.
+ */
+
+namespace App\Fetcher;
+
+interface FetcherInterface
+{
+    public function getUrl(array $params): string;
```
welcoMattic: $params should be optional. Not all fetcher needs parameters.

src/Entity/Conference.php
```diff
@@ -76,7 +76,7 @@ class Conference
     private $startAt;
 
     /**
-     * @ORM\Column(name="end_at", type="datetime")
+     * @ORM\Column(name="end_at", type="datetime", nullable=true)
```
welcoMattic: 🤔 a conference must have an end date, even if it's the same as the start date (one day conference)

src/Fetcher/ConfTechFetcher.php
```diff
@@ -32,136 +34,206 @@ public function __construct(RegistryInterface $doctrine, LoggerInterface $logger
...
-            ],
-        ];
+        $params = $this->matchTags();
 
         $conferencesFinal = [];
+        $newConferencesCount = 0;
 
         foreach ($params as $date => $technologies) {
             foreach ($technologies as $technologie) {
                 $client = new Client();
```
welcoMattic: The Client should be initialized outside of the loops, to avoid reinitialization

src/Fetcher/ConfTechFetcher.php
```diff
@@ -32,136 +34,206 @@ public function __construct(RegistryInterface $doctrine, LoggerInterface $logger
...
+    private function matchTags()
     {
-        $location = $conference->city;
+        $dateCurrentYear = date('Y');
+        $dateNextYear = date('Y', strtotime('+1 year'));
+
+        $tagsSelected = $this->em->getRepository(Tag::class)->findBy(['selected' => true]);
+
+        $tagMatch = [
+            'Android' => 'android',
```
welcoMattic: For keys you should use Enum instead of hard rewrite tags.

src/Fetcher/ConfTechFetcher.php
```diff
@@ -32,136 +34,206 @@ public function __construct(RegistryInterface $doctrine, LoggerInterface $logger
...
+    }
+
+    private function getTagByName(string $tagName): Tag
+    {
+        $tag = $this->em->getRepository(Tag::class)->findOneBy([
+            'name' => $tagName,
+        ]);
+
+        return $tag;
+    }
```
welcoMattic: This method should be placed in the TagRepository class

src/Fetcher/ConfTechFetcher.php
```diff
@@ -32,136 +34,206 @@ public function __construct(RegistryInterface $doctrine, LoggerInterface $logger
...
+
+    private function setParams($fC)
```
welcoMattic: This method is not well named IMO. I will call it `hydrateConference`

src/Entity/Conference.php
alexzerah: Some conferences do not have an end date (2 for conf-tech source  for example).
If so, do I put the end date = start date?

welcoMattic: Ok. No, leave `null` as value

src/Fetcher/ConfTechFetcher.php
alexzerah: How do I know which key is it ?
Do I need to keep the same order than Tag enum ? Or in alphabetical order for both ?

welcoMattic: You can do `array_combine(TagEnum::toArray(), $confTechFetcherSynonyms)` where `confTechFetcherSynonyms` contains all values of `$tagMatch` array. In this case, yes the order matters.

alexzerah:
Currently, the fetcher command allows to :
- fetch data from class which implements FetcherInterface
- Add conferences in database
- Check if conference is already in database with hash
- get arrays of conf 

I still have to :
- Update some conference fields if changed (site_url, cfp_url, ...). Now, if a conference has been added, it's ignored.
- Change slug to put year (to avoid "front-end-ne" and "front-end-ne-1")

alexzerah:
After performing a review, this is the state of the PR :
- All requested changes have been done ☑
- The PR fixes all issue to-do execept one. The last one need to be validate once the following is done:
  - Allow editing of fields in database for conferences (currently when one is added, it can't be updated)
  - Update conférence slug (some conferences have same name) 
- Some test is not valid. This is why CircleCI failed. This doesn't seem to be due to the PR but to the master branch. The error is `An exception occurred in driver: SQLSTATE[HY000] [2002] No such file or directory` (CircleCI show a HTML response). The test fail seems due to a wrong database configuration.

I think this changes can be done in 1-2 hours. 

 @welcoMattic what do you think about it ?

src/Fetcher/ConfTechFetcher.php
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    public function fetch(): array
+    {
+        $params = $this->matchTags();
+
+        $conferences = [];
+        $client = new Client();
+
+        foreach ($params as $date => $technologies) {
```
welcoMattic: You should replace `$technologies` by `$tags` to be consistent

src/Fetcher/ConfTechFetcher.php
```diff
@@ -0,0 +1,209 @@
...
+            if (null !== $tagMatch[$tag->getName()]) {
+                array_push($params[$dateCurrentYear], [$tagMatch[$tag->getName()], $tag->getName()]);
+                array_push($params[$dateNextYear], [$tagMatch[$tag->getName()], $tag->getName()]);
+            }
+        }
+
+        return $params;
+    }
+
+    private function hydrateConference($fC)
```
welcoMattic: This method could be replaced by a call to Symfony Serializer with a custom Normalizer

src/Fetcher/ConfTechFetcher.php
```diff
@@ -0,0 +1,209 @@
...
+
+        if (isset($fC->cfpEndDate)) {
+            $cfpEndAt = \DateTime::createFromFormat('Y-m-d h:i:s', $fC->cfpEndDate.' 00:00:00');
+            $conference->setCfpEndAt($cfpEndAt);
+        }
+
+        return $conference;
+    }
+
+    private function hash(object $fC)
```
welcoMattic: This method must be extracted to an isolated service, because we will need it in all Fetchers

src/Repository/ConferenceRepository.php
```diff
@@ -63,6 +63,17 @@ public function findOneAttended(string $slug): ?Conference
...
+    public function findOneByHash(string $hash): ?Conference
+    {
+        $qb = $this->createQueryBuilder('c')
+            ->where('c.hash = :hash')
+            ->setParameter('hash', $hash)
+            ->getQuery()
+            ->getOneOrNullResult();
+
+        return $qb;
+    }
```
welcoMattic: Doctrine has `findOneBy` method, which is allow you to find one entity by the first argument. i.e: `$repo->findOneBy('hash', $hash)`. So, you could remove this method

src/Repository/TagRepository.php
```diff
@@ -0,0 +1,57 @@
...
+            ->setParameter('name', $tagName)
+            ->getQuery();
+        try {
+            $qb = $qb->getSingleResult();
+        } catch (\Exception $e) {
+            $this->logger->error($e);
+        }
+
+        return $qb;
+    }
```
welcoMattic: Doctrine has findOneBy method, which is allow you to find one entity by the first argument. i.e: $repo->findOneBy('name', $name). So, you could remove this method

src/Repository/TagRepository.php
```diff
@@ -0,0 +1,57 @@
...
+
+    public function getTagsBySelected(): array
+    {
+        return $qb = $this->createQueryBuilder('t')
+            ->where('t.selected = :selected')
+            ->setParameter('selected', true)
+            ->getQuery()
+            ->execute();
+    }
```
welcoMattic: idem with selected

welcoMattic:
@alexzerah you can review it if you want ;)

PR #46, opened 2019-07-18
Fix slack message in double

### LINKED ISSUE #44: Fix double Slack message for added conferences
Sometimes we have Slack message in double for new added conferences. We need to investigate to find why and fix it.

In otherwise, we also have "empty" Slack message with only CFP link as CFP reminder message. May be it is because we do not have conference data? Needs investigation too.

welcoMattic:
Fix #44

### DIFF
--- a/src/Event/CfpEndingSoonEvent.php
+++ b/src/Event/CfpEndingSoonEvent.php
@@ -41,25 +41,26 @@ class CfpEndingSoonEvent extends Event
         $cfpAttachment = SlackNotifier::ATTACHMENT;
         $template = '🔊  CFP for %s (%s) is closing %s';
         $conferenceLink = sprintf('<%s|%s>', $this->conference->getCfpUrl(), $this->conference->getName());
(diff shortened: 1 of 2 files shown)


src/Event/CfpEndingSoonEvent.php
```diff
@@ -41,25 +41,26 @@ public function buildAttachment(): array
         $cfpAttachment = SlackNotifier::ATTACHMENT;
         $template = '🔊  CFP for %s (%s) is closing %s';
         $conferenceLink = sprintf('<%s|%s>', $this->conference->getCfpUrl(), $this->conference->getName());
+        $countdown = 'in *' . $this->remainingDays . ' day' . $this->remainingDays > 1 ? 's' : '' . '* !';
```
Korbeil: et les pluriels hein ? :stuck_out_tongue:

welcoMattic: Yes, it's mentioned in #47

Issue #29, opened 2019-05-09
Implement other login methods


welcoMattic:
- Add a classic email/password method
- Add a GitHub login (with organization restriction if possible)

PR #49, opened 2019-07-18
Add Form login and GitHub login

welcoMattic:
Ref #29 

Todo:

- [x] Add login restriction based on GitHub Ogranizations list from user settings.


welcoMattic:
Tests are green, CI is broken due to infrastructure update (python 2.7 to python 3.8). It seems CircleCi has some cache which is not cleared, and not clearable from the UI.

Issue #57, opened 2019-12-05
Add a whitelist/blacklist system to filter more precisely conference fetching


welcoMattic:
The actual fetching system is not ideal. We fetch all conferences based on a tags list. But for instance at JoliCode we fetch all PHP conference, including TYPO3 conference, which is a CMS that we do not use. So, we would like to manage a blacklist of tags and words contained in conferences name to avoid fetching them.

The behaviour should be:
We fetch all conferences based on the whitelist of tags, and we remove all matching conferences on the blacklist, then we record the result of this filtering.

PR #59, opened 2019-12-05
Add an exclusion system to filter conference fetching

remyvanlerberghe:
Ref #57

### DIFF
--- a/src/Command/RemindEndingCfpCommand.php
+++ b/src/Command/RemindEndingCfpCommand.php
@@ -49,7 +49,7 @@ class RemindEndingCfpCommand extends Command
         foreach ($conferences as $conference) {
             $remainingDays = (int) ($conference->getCfpEndAt()->diff($today)->format('%a'));
 
-            if (\in_array($remainingDays, self::REMAINING_DAYS_STEPS, true) && !$conference->getExcluded()) {
+            if (\in_array($remainingDays, self::REMAINING_DAYS_STEPS, true)) {
                 $this->eventDispatcher->dispatch(new CfpEndingSoonEvent($conference, $remainingDays));
             }
         }
--- a/src/Command/RemindEndingCfpCommand.php
+++ b/src/Command/RemindEndingCfpCommand.php
@@ -49,7 +49,7 @@ class RemindEndingCfpCommand extends Command
         foreach ($conferences as $conference) {
             $remainingDays = (int) ($conference->getCfpEndAt()->diff($today)->format('%a'));
 
-            if (\in_array($remainingDays, self::REMAINING_DAYS_STEPS, true)) {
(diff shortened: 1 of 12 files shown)


src/Command/RemindEndingCfpCommand.php
```diff
@@ -49,7 +49,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         foreach ($conferences as $conference) {
             $remainingDays = (int) ($conference->getCfpEndAt()->diff($today)->format('%a'));
 
-            if (\in_array($remainingDays, self::REMAINING_DAYS_STEPS, true)) {
+            if (\in_array($remainingDays, self::REMAINING_DAYS_STEPS, true) && !$conference->getExcluded()) {
```
welcoMattic: It could be better to filter directly at SQL level, we do not need to fetch from database excluded conferences.

src/Fetcher/ConfTechFetcher.php
```diff
@@ -146,6 +150,14 @@ public function denormalizeConferences(array $rawConferences, string $source, Ta
             $conference->setSiteUrl($rawConference['url']);
             $conference->addTag($tag);
 
+            $excluded = false;
+            foreach ($this->excludedTags as $excludedTag) {
+                if (fnmatch($excludedTag->getName(), $rawConference['name'], FNM_CASEFOLD)) {
+                    $excluded = true;
```
welcoMattic: You can break the loop after this because it will never happen to reset `$excluded` to false.

src/Fetcher/JoindApiFetcher.php
```diff
@@ -149,6 +153,13 @@ public function denormalizeConferences(array $rawConferences, string $source, Ta
             $conference->setSiteUrl($rawConference['href']);
             $conference->addTag($tag);
 
+            $excluded = false;
+            foreach ($this->excludedTags as $excludedTag) {
+                if (fnmatch($excludedTag->getName(), $rawConference['name'], FNM_CASEFOLD)) {
+                    $excluded = true;
```
welcoMattic: idem

src/Repository/ConferenceRepository.php
```diff
@@ -33,6 +33,7 @@ public function findEndingCfps(): array
 
         return $this->createQueryBuilder('c')
             ->andWhere('c.cfpEndAt IS NOT NULL AND c.cfpEndAt >= :today AND c.cfpEndAt < :threshold')
+            ->andWhere('c.excluded = FALSE')
```
welcoMattic:
```suggestion
            ->andWhere('c.excluded = :excluded')
            ->setParameter('excluded', false)
```

We let Doctrine casts the value itself

welcoMattic:
Need to be rebased on master

Issue #58, opened 2019-12-05
Add region filter on fetching


welcoMattic:
It could be useful to avoid fetching conference placed on the other side of the world. So, we want to filter conference fetching on region (by continent, by country). For fetchers without this filter possibility, we want to filter after fetching, before persist on database.

PR #60, opened 2019-12-06
Add continent filter on fetching

remyvanlerberghe:
Ref #58


src/Fetcher/ContinentGuesser.php
```diff
@@ -0,0 +1,45 @@
...
+use Geocoder\StatefulGeocoder;
+use Http\Adapter\Guzzle6\Client;
+use SameerShelavale\PhpCountriesArray\CountriesArray;
+use Symfony\Bridge\Doctrine\RegistryInterface;
+
+class ContinentGuesser
+{
+    private $continents;
+
+    public function __construct(RegistryInterface $doctrine)
```
welcoMattic: I prefer to inject directly the repository here

welcoMattic:
Need to be rebased on master

Issue #61, opened 2019-12-06
Can't access Participation list page


remyvanlerberghe:
After upgrade EasyAdmin from `^2.1` to `^2.3`, this [Pull Request](https://github.com/EasyCorp/EasyAdminBundle/pull/2692) introduces a Breaking Change and results on a 500 Error on Participation list page

PR #62, opened 2019-12-06
Fixed easyadmin list actions for Participation

remyvanlerberghe:
Ref #61


welcoMattic:
Cs fix will be corrected in a future PR

Issue #76, opened 2020-06-25
Find a way to avoid duplicated conferences


welcoMattic:
For now, we have 2 conferences sources (JoindIn API and Conf Tech JSON files). 
Sometimes, we fetch the same conference from each source, but the name is not exactly the same.

For instance, "Forum PHP" in Conf Tech, and "Forum PHP 2020" in JoindIn API.
Both has the same dates, but not the same name, and not the same CFP url (https://afup.org/event/forumphp2020 and http://cfp.afup.org/).

To avoid this, @lyrixx gives us an hint: https://www.php.net/manual/en/function.levenshtein.php

We could match the incoming conferences to the existing ones with `levenshtein` and avoid storing multiple times the same conference.

PR #77, opened 2020-07-02
Fix duplicated conference on fetch

welcoMattic:
Ref https://github.com/jolicode/starfleet/issues/76

Conferences are fetched from different source, and sometimes they do not have exactly the same name (with or without year, or different case, or even a typo inside).

This PR fixes the problem by using levenshtein algorithm to compute the distance between fetched conferences and existing ones.
If the distance is 0, and the start and end date are equals, then the conferences are the same.
If not, at the end of the process we get the shortest distance and check if it's lower than 3 and start and end date are the same, if so then the conferences are the same. We may be need to adjust this one in the future.


src/Command/FetchConferencesCommand.php
```diff
@@ -43,55 +45,98 @@ protected function configure()
         $this->setDescription('Fetch conferences from Fetcher Classes');
     }
 
-    protected function execute(InputInterface $input, OutputInterface $output)
+    /**
+     * @param InputInterface $input
+     * @param ConsoleOutputInterface $output
+     * @return int
+     */
```
Korbeil, line 52: Pourquoi ? Le prototype de ta méthode est déjà bien typé.

src/Command/FetchConferencesCommand.php
```diff
@@ -43,55 +45,98 @@ protected function configure()
...
-            $this->em->flush();
+        if ($shortest > 3) {
+            $matchedConference = null;
         }
 
-        $symfonyStyle->writeln("\n");
-        $symfonyStyle->success($newConferencesCount.' newly added conference(s)');
-        $symfonyStyle->success($updatedConferencesCount.' updated conference(s)');
+        return $matchedConference;
     }
```
Korbeil, line 140: C'est une bonne idée, mais je pense qu'il faut pas se limiter au nom de la conférence, les dates devraient aussi êtres prises en compte pour vérifier si on a déjà la conf ou non.
Comme typiquement, pour le nom, entre "Forum PHP 2020" et "Forum PHP 2021" la distance est très faible, alors que les dates sont complètement différentes.

src/Command/FetchConferencesCommand.php
welcoMattic: Not exactly, I use `ConsoleOutputInterface` in PHPDoc, which allows completion for `section` method

Issue #85, opened 2020-11-25
Improve Talks list view on admin


welcoMattic:
The actual talks list view is very "empty"
![image](https://user-images.githubusercontent.com/773875/100237901-d47f1a00-2f2f-11eb-8944-ec275cafd646.png)

It should be great to add some columns like "authors", "created at", … In this way, it will be possible to filter on these fields to easily retrieve Talks.

lyrixx:
We **might** use https://github.com/alterphp/EasyAdminExtensionBundle

PR #88, opened 2020-12-02
Admin view

HedicGuibert:
This PR fixes [#85](https://github.com/jolicode/starfleet/issues/85#issue-750847364) .

It adds createdAt and speakers columns to the admin talk view. It also adds a talks column to the user list view, but this may be unnecessary.

Talks list view :
Talks

Users list view :
EDIT :  removed

Users

### DIFF
--- a/config/packages/easy_admin/talk.yaml
+++ b/config/packages/easy_admin/talk.yaml
@@ -8,7 +8,7 @@ easy_admin:
                     - { property: 'title', type: 'text' }
                     - { property: 'submits' }
                     - { property: 'createdAt', type: 'date' }
-                    - { property: 'authors', type: 'text', template: 'easy_admin/talk/authors.html.twig' }
+                    - { property: 'authors', template: 'easy_admin/talk/authors.html.twig' }
             show:
                 fields:
                     - { property: 'title' }
--- a/config/packages/easy_admin/talk.yaml
+++ b/config/packages/easy_admin/talk.yaml
@@ -8,7 +8,7 @@ easy_admin:
                     - { property: 'title', type: 'text' }
                     - { property: 'submits' }
                     - { property: 'createdAt', type: 'date' }
-                    - { property: 'authors', type: 'string', template: 'easy_admin/Talk/authors.html.twig' }
+                    - { property: 'authors', type: 'text', template: 'easy_admin/talk/authors.html.twig' }
             show:
                 fields:
                     - { property: 'title' }
--- a/config/packages/easy_admin/talk.yaml
+++ b/config/packages/easy_admin/talk.yaml
@@ -7,6 +7,8 @@ easy_admin:
                 fields:
                     - { property: 'title', type: 'text' }
                     - { property: 'submits' }
+                    - { property: 'createdAt', type: 'date' }
+                    - { property: 'authors', type: 'string', template: 'easy_admin/Talk/authors.html.twig' }
             show:
                 fields:
                     - { property: 'title' }
--- a/src/DataFixtures/TalkFixtures.php
+++ b/src/DataFixtures/TalkFixtures.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace App\DataFixtures;
+
+use App\Entity\Submit;
+use App\Entity\Talk;
+use App\Entity\User;
+use Doctrine\Bundle\FixturesBundle\Fixture;
+use Doctrine\Persistence\ObjectManager;
+
+class TalkFixtures extends Fixture
+{
+    public function load(ObjectManager $manager)
+    {
+
+        $talk = new Talk();
+        $talk->setTitle('TalkTitle');
+        $talk->setIntro('TalkIntro');
+        $talk->addAuthor($this->getReference('user'));
+        $talk->addAuthor($this->getReference('admin'));
+        $manager->persist($talk);
+
+        $manager->flush();
+    }
+}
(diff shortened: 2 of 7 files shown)


lyrixx:
Please, add some screenshot. Thanks

HedicGuibert:
Sure, sorry for this omission.
I updated my innitial comment.

welcoMattic:
I'm afraid that the talks column on the user list view is not necessary. In fact, the talk list of an user will increase over time, so the content of this column to, it will quickly become unreadable.

src/DataFixtures/TalkFixtures.php
```diff
@@ -0,0 +1,25 @@
...
+use App\Entity\Talk;
+use App\Entity\User;
+use Doctrine\Bundle\FixturesBundle\Fixture;
+use Doctrine\Persistence\ObjectManager;
+
+class TalkFixtures extends Fixture
+{
+    public function load(ObjectManager $manager)
+    {
+
```
lyrixx, line 15: useless space

HedicGuibert:
Ok, this makes sense, I'm removing it then

src/Entity/Talk.php
```diff
@@ -46,9 +46,15 @@ class Talk
      */
     private $submits;
 
+    /**
+     * @ORM\OneToMany(targetEntity=User::class, mappedBy="talks")
+     */
+    private $authors;
```
lyrixx, line 52: Why did you add this new property? more over this seems wrong (never updated, never used)

lyrixx (changes requested):

src/Entity/Talk.php
welcoMattic, line 52: Hmm, I think there is a design issue here. Actually the data schema is:
User 1->n Submit n->1 Talk
Here, you are adding a direct relation between User and Talk, but this relation exists, indirectly through Submit. 
So, you might be able to retrieve all talks of a user through there submits, with something like:

```
    public function getAuthors(): string
    {
        return array_reduce($this->getSubmits(), function ($r, Submit $submit) {
            foreach ($submit->getUsers() as $user) {
                $r = '' === $r ? $user->getName() : $r . ', ' . $user->getName();
            }

            return $r;
        }, '');
    }
```

(I did not test this code)

HedicGuibert: Isn't a relation needed between Talk and User to allow the speakers to be displayed on the talks list view ?

HedicGuibert: Oh sure, it looks a lot better indeed and adresses what is said above.
Thanks !

lyrixx: And BTW, the code here is ugly. Use a template in easy admin for that

templates/easy_admin/Talk/authors.html.twig
```diff
@@ -0,0 +1,5 @@
+{% for submit in value %}
+    {% for user in submit.users %}
```
lyrixx, line 2: I would use `<ul><li>` here ?

config/packages/easy_admin/talk.yaml
```diff
@@ -7,6 +7,8 @@ easy_admin:
                 fields:
                     - { property: 'title', type: 'text' }
                     - { property: 'submits' }
+                    - { property: 'createdAt', type: 'date' }
+                    - { property: 'authors', type: 'string', template: 'easy_admin/Talk/authors.html.twig' }
```
lyrixx, line 11: `Talk` -> `talk`

src/Entity/Talk.php
```diff
@@ -101,6 +101,11 @@ public function removeSubmit(Submit $submit): self
         return $this;
     }
 
+    public function getAuthors()
+    {
+        return $this->getSubmits();
```
lyrixx, line 106: The wording seems odd Author VS Submits, and anyway, I don't think this method is useful anyway :)

lyrixx:
Could you also sqash your commits, because it does not say what they do

config/packages/easy_admin/talk.yaml
```diff
@@ -7,6 +7,8 @@ easy_admin:
                 fields:
                     - { property: 'title', type: 'text' }
                     - { property: 'submits' }
+                    - { property: 'createdAt', type: 'date' }
+                    - { property: 'authors', type: 'text', template: 'easy_admin/talk/authors.html.twig' }
```
lyrixx, line 11: The type here is useless since you are using a template

templates/easy_admin/talk/authors.html.twig
```diff
@@ -0,0 +1,7 @@
+{% for submit in item.submits %}
+    <ul style="margin-bottom:0">
```
lyrixx, line 2: I would put the `ul` before the first for, or you'll get many ul in this table cell.
And IMHO, we will have lot's of duplicate content here.
If a user submit 5 talks. He'll appear here 5 times.
You should unique the list

welcoMattic: Even if most of time there will be 1 speaker, if a talk is submitted 2 times with 2 different speakers couple (let say Alice and Bob, and Alice and Jack), we will see "Alice, Bob, Jack", which is not very relevant, don't you think? 
May be unique the list, but "by submit"
1st submit: Alice
2nd submit: Alice
3rd submit: Alice and Bob
4th submit: Alice and Jack
5th submit: Alice and Bob
will result to:
Alice, Alice + Bob, Alice + Jack

WDYT?

HedicGuibert:
I updated the view accordingly to what you said, and I added some fixtures.
Here is what it looks like now :
NewTalksView

lyrixx (approved):
I approve, even if you did not uniq the list. That's an edge and it seems you did not manage to succeed. Not a big deal.

HedicGuibert:
I did it locally but I felt like it wasn't useful, seems I was wrong ? I can still add it

PR #89, opened 2020-12-03
Add missing headers

HedicGuibert:
So I forgot to cs-fix in previous PR and the headers are missing in the fixtures I created.

This simply adds them


lyrixx (approved):
You can merge when green 👍🏼

PR #90, opened 2020-12-03
Bugfix - talks list view  

HedicGuibert:
This fixes a bug where, if a user was giving the same talk multiple times alone, it would add a / after his name even if he's alone in his column.
This was caused by the `{% if not loop.last %}/{% endif %}` and the unique filtering we apply. Twig sees there are 2 submits, so loops twice and adds a / at the end of the first, which is not end of loop. But it prevents the second from being displayed because the user is not unique so it just leaves the /.

So we send submits with the same author alone only once.

EDIT : Ok so I ended up doing quite a bit more than fixing a bug. The bug is fixed by making the code way simpler, and it also rewrites the fixtures and add tests. It should be way better now.


tests/TalkTest.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace App\Tests\Entity;
+
+use App\Entity\Submit;
+use App\Entity\Talk;
+use PHPUnit\Framework\TestCase;
+
+class TalkTest extends TestCase
```
lyrixx, line 9: This file is not in the right directory. It should follow the same convention as the `src/` dir

lyrixx (commented):
This is a very good start 👍🏼

HedicGuibert:
I updated and squashed to make it less messy.

src/Entity/Talk.php
```diff
@@ -101,6 +101,24 @@ public function removeSubmit(Submit $submit): self
         return $this;
     }
 
+    public function getUniqueUsers()
```
lyrixx, line 104: missing type hint

templates/conferences/list.html.twig
```diff
@@ -69,7 +69,7 @@
                 </div>
 
                 {% set activeTab = activeTab|default(liveConferences|length > 0 ? 'live' : futureConferences|length > 0 ? 'future' : 'past') %}
-
+
```
lyrixx, line 72: oups

tests/Entity/TalkTest.php
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+            $talk->addSubmit($newSubmit);
+        }
+
+        $result = $talk->getUniqueUsers();
+        self::assertSame($expectedAmount, \count($result));
+    }
+
+    public function provideUsers()
```
lyrixx, line 45: missing type hint (`Generator`)

tests/Entity/TalkTest.php
```diff
@@ -0,0 +1,66 @@
...
+                $newUser->setName($user);
+
+                $newSubmit->addUser($newUser);
+            }
+
+            $talk->addSubmit($newSubmit);
+        }
+
+        $result = $talk->getUniqueUsers();
+        self::assertSame($expectedAmount, \count($result));
```
lyrixx, line 42: There is a method in phpunit for such assertion (`assetCount`)

lyrixx (commented):
Except some minor comments, this is nice 👍🏼 Well done

lyrixx (approved):

PR #94, opened 2020-12-16
Add flags to conferences

HedicGuibert:
This PR addresses #3.

It adds a country and city fields to the Conference entity rather than a location field, and it displays them with flag in the nextConference adminView.

How it looks on the frontend :
Frontend flag

How it looks in the admin :
Flags View

How to add a new country :
Save Conf

What it looks like now:
Flags View2


src/Entity/Conference.php
```diff
@@ -60,11 +60,6 @@ class Conference
      */
     private $description;
 
-    /**
-     * @ORM\Column(name="location", type="string", length=255)
-     */
-    private $location;
-
```
welcoMattic, line 67: Be careful here, we will lost data stored in this field. Be sure that your migration will move data from Location to City or Country fields

migrations/Version20201216112814.php
```diff
@@ -0,0 +1,44 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20201216112814 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->abortIf('postgresql' !== $this->connection->getDatabasePlatform()->getName(), 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('DROP INDEX uniq_911533c8d1b862b8');
+        $this->addSql('ALTER TABLE conference DROP hash');
```
welcoMattic, line 26: Why `hash` column is dropped here? It looks like an mistake

migrations/Version20201216112814.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->addSql('ALTER TABLE conference RENAME COLUMN location TO city');
```
welcoMattic, line 27: Nice :ok_hand:

welcoMattic:
@HedicGuibert We want to display the flag on the frontend side too

migrations/Version20201216112814.php
HedicGuibert: I agree, I don't know what it's doing here, I saw it and thought it may have been done previously and not added to a migration so I let it here, but I can drop it :+1:

migrations/Version20201216112814.php
```diff
@@ -0,0 +1,44 @@
...
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20201216112814 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->abortIf('postgresql' !== $this->connection->getDatabasePlatform()->getName(), 'Migration can only be executed safely on \'postgresql\'.');
```
lyrixx, line 23: Should be removed (I hate theses checks :))

migrations/Version20201216112814.php
```diff
@@ -0,0 +1,44 @@
...
+
+        $this->addSql('DROP INDEX uniq_911533c8d1b862b8');
+        $this->addSql('ALTER TABLE conference DROP hash');
+        $this->addSql('ALTER TABLE conference RENAME COLUMN location TO city');
+        $this->addSql('ALTER TABLE conference ADD country VARCHAR(255) DEFAULT \'Undefined\' NOT NULL');
```
lyrixx, line 28: I don't like this "Undefined". Let's use `null` is done for that

migrations/Version20201216112814.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->addSql('ALTER TABLE users ALTER email TYPE VARCHAR(255)');
```
lyrixx, line 29: This is suspect

migrations/Version20201216112814.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->abortIf('postgresql' !== $this->connection->getDatabasePlatform()->getName(), 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE SCHEMA public');
```
lyrixx, line 36: To remove

src/Entity/Conference.php
```diff
@@ -116,6 +111,16 @@ class Conference
      */
     private $participations;
 
+    /**
+     * @ORM\Column(type="string", length=255)
+     */
+    private $country;
```
lyrixx, line 117:
```suggestion
    private ?string $country = null;
```

same after. We are using PHP 7.4, let's use it :)

lyrixx (commented):
I don't see the code that display flags
+ I don't see where you set the city / location
+ Could you add a screenshot?

migrations/Version20201216112814.php
HedicGuibert: I could but the field is not nullable. Should I set it to Nullable then ? Or maybe this will work work anyway ?

lyrixx: if the country could be missing, so yes the field must be nullable

migrations/Version20201216112814.php
lyrixx: I merged another PR where he fixed this stuff :)

lyrixx:
@HedicGuibert Could you rebase to get ride of not-related SQL diff? thanks

HedicGuibert:
I updated the description and made the requested changes.
@welcoMattic You may want to look closely at the NewConferenceEvent, since I added a new attachment that may not be right.

welcoMattic:
@HedicGuibert Looks good, in a Slack block you can add many attachments, no worries about adding one for City ;)

lyrixx:
Hi, I don't understand how this PR can work.
Basically, you renamed `setLocation` to `setCity` and `setCountry`.
I saw you change things in the fixtures and the admin. This is nice.

But I did not see any change in the frontend, nor in the fetcher, etc
=> How could it continue to work ?
=> Do the command to fetch conference continue to work ?
You need to search in the code for every single usage of get/setLocation() and update them.

config/services.yaml
```diff
@@ -49,3 +49,18 @@ services:
     App\Security\Authenticator\GitHubAuthenticator:
         arguments:
             $allowedGitHubOrganization: '%env(APP_ALLOWED_GITHUB_ORGANIZATION)%'
+
+    client:
```
lyrixx, line 53: this name is too generic / vague. Please use something like `http.guzzle.client`

lyrixx: Nevermind: use the classname for the ie, same after

lyrixx (commented):
Except one little comment, this seems good !

lyrixx:
I'll wait for PHP before merging this one

lyrixx:
Thanks, good job 👍🏼

PR #96, opened 2020-12-16
Add forgotten fixture setter

HedicGuibert:
I forgot a setter in the fixtures leading to `null` in the cfpEndAt column in the nextConference list view


lyrixx (approved):
👍🏼 

I don't like the fact that the CFP and the conference ends at the same date, but we can change it later if needed.

You can auto merge when green

PR #99, opened 2020-12-17
Fix minor issues in fixtures and talk list view

HedicGuibert:
This addresses two mnior issues :

-  margin bottom on `ul` in talk list view was not great
-  cfpEndAt was set to endAt in the default fixtures


lyrixx (approved):
You can merge when green

PR #100, opened 2020-12-18
Add PHPstan

HedicGuibert:
This adds phpstan and the fixes requested by it.
It also upgrades to 5.2.1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
                   when: always
             - run:
                     name: Run PHPStan
-                    command: php -d memory_limit=-1 vendor/bin/phpstan analyse -c ./phpstan.neon.dist src
+                    command: php -d memory_limit=-1 vendor/bin/phpstan analyse
 
     tests:
         machine:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,6 +50,9 @@ jobs:
                   name: Lint Twig
                   command: php bin/console lint:twig templates/
                   when: always
+            - run:
+                    name: Run PHPStan
+                    command: php -d memory_limit=-1 vendor/bin/phpstan analyse -c ./phpstan.neon.dist src
 
     tests:
         machine:
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,5 +1,6 @@
 includes:
     - vendor/phpstan/phpstan-doctrine/extension.neon
+    - vendor/phpstan/phpstan-doctrine/rules.neon
     - vendor/phpstan/phpstan-symfony/extension.neon
 
 parameters:
@@ -13,4 +14,5 @@ parameters:
     symfony:
         container_xml_path: 'var/cache/dev/App_KernelDevDebugContainer.xml'
 
-    ignoreErrors: []
+    doctrine:
+        objectManagerLoader: var/phpstan-utilities/object-manager.php
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,0 +1,16 @@
+includes:
+    - vendor/phpstan/phpstan-doctrine/extension.neon
+    - vendor/phpstan/phpstan-symfony/extension.neon
+
+parameters:
+    level: 5
+    paths:
+        - src
+        
+    tmpDir: var/phpstan-tmp
+    inferPrivatePropertyTypeFromConstructor: true
+
+    symfony:
+        container_xml_path: 'var/cache/dev/App_KernelDevDebugContainer.xml'
+
+    ignoreErrors: []
--- a/src/Command/RemindEndingCfpCommand.php
+++ b/src/Command/RemindEndingCfpCommand.php
@@ -13,6 +13,7 @@ namespace App\Command;
 
 use App\Entity\Conference;
 use App\Event\CfpEndingSoonEvent;
+use App\Repository\ConferenceRepository;
 use Doctrine\Persistence\ManagerRegistry;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
@@ -23,14 +24,16 @@ class RemindEndingCfpCommand extends Command
 {
     const REMAINING_DAYS_STEPS = [30, 20, 10, 5, 1, 0];
 
-    private $repository;
+    private ConferenceRepository $repository;
     private $eventDispatcher;
 
     public function __construct(ManagerRegistry $doctrine, EventDispatcherInterface $eventDispatcher)
     {
         parent::__construct();
 
-        $this->repository = $doctrine->getManager()->getRepository(Conference::class);
+        /** @var ConferenceRepository $conferenceRepository */
+        $conferenceRepository = $doctrine->getManager()->getRepository(Conference::class);
+        $this->repository = $conferenceRepository;
         $this->eventDispatcher = $eventDispatcher;
     }
 
@@ -40,7 +43,7 @@ class RemindEndingCfpCommand extends Command
         $this->setDescription('Trigger event for cfp ending soon');
     }
 
-    protected function execute(InputInterface $input, OutputInterface $output)
+    protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $today = new \DateTime();
         $today->setTime(0, 0, 0);
@@ -53,5 +56,7 @@ class RemindEndingCfpCommand extends Command
                 $this->eventDispatcher->dispatch(new CfpEndingSoonEvent($conference, $remainingDays));
             }
         }
+
+        return 0;
     }
 }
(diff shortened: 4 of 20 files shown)


tests/Functional/CreateAuthenticatedClientTrait.php
```diff
@@ -25,7 +25,7 @@ protected static function createAuthenticatedClient(User $user = null, bool $isA
         if (!$user) {
             $email = ($isAdmin ? 'admin' : 'user').'@starfleet.app';
             $user = $container->get('doctrine')->getRepository(User::class)
-                ->findOneByEmail($email);
+                ->__call('findOneByEmail', [$email]);
```
welcoMattic, line 28:
```suggestion
                ->findOneBy(['email' => $githubUser->getEmail()]);
```

welcoMattic (changes requested):

.circleci/config.yml
```diff
@@ -50,6 +50,9 @@ jobs:
                   name: Lint Twig
                   command: php bin/console lint:twig templates/
                   when: always
+            - run:
+                    name: Run PHPStan
+                    command: php -d memory_limit=-1 vendor/bin/phpstan analyse -c ./phpstan.neon.dist src
```
lyrixx, line 55: src is uselss here since it's already in the config file. 
BTW, the config file option is useless too, since it's the default value
And Renamed the config file to `phpstan.neon`

phpstan.neon.dist
```diff
@@ -0,0 +1,16 @@
...
+    level: 5
+    paths:
+        - src
+        
+    tmpDir: var/phpstan-tmp
+    inferPrivatePropertyTypeFromConstructor: true
+
+    symfony:
+        container_xml_path: 'var/cache/dev/App_KernelDevDebugContainer.xml'
+
```
lyrixx, line 15: I don't see the configuration for doctrine

phpstan.neon.dist
```diff
@@ -0,0 +1,16 @@
...
+    ignoreErrors: []
```
lyrixx, line 16: useless code is useless :) To remove

src/Command/RemindEndingCfpCommand.php
```diff
@@ -23,14 +24,16 @@ class RemindEndingCfpCommand extends Command
...
+    private ConferenceRepository $repository;
     private $eventDispatcher;
 
     public function __construct(ManagerRegistry $doctrine, EventDispatcherInterface $eventDispatcher)
     {
         parent::__construct();
 
-        $this->repository = $doctrine->getManager()->getRepository(Conference::class);
+        /** @var ConferenceRepository $conferenceRepository */
+        $conferenceRepository = $doctrine->getManager()->getRepository(Conference::class);
```
lyrixx, line 35: This is a bad practice to inject the Entity Manager to get ta Repository. Inject the repository instead.
It's a bad practice because
* It hides a deps
* class become harder to test
(same for Symfony Container)

src/Security/Authenticator/GitHubAuthenticator.php
```diff
@@ -80,11 +81,11 @@ public function getUser($credentials, UserProviderInterface $userProvider): ?Use
             return $existingUser;
         }
 
-        /** @var User $user */
+        /** @var User|null $user */
         $user = $this->em->getRepository(User::class)
             ->findOneBy(['email' => $githubUser->getEmail()]);
 
-        if ($user) {
+        if ($user instanceof User) {
```
lyrixx, line 88: I hate this kind of change. Previous code was totally valid.
Is it mandatory?

src/Security/Authenticator/GitHubAuthenticator.php
```diff
@@ -115,7 +116,9 @@ public function start(Request $request, AuthenticationException $authException =
 
     public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
     {
-        $this->requestStack->getCurrentRequest()->getSession()->getFlashBag()->add('error', 'Your GitHub user is not in the allowed organization, or the membership is private.');
+        /** @var Session $session */
+        $session = $this->requestStack->getCurrentRequest()->getSession();
```
lyrixx, line 120: * `$request` can be null
* this kind of change brings totally nothing to the project

lyrixx (commented):
Except few comments, this seems fine

src/Security/Authenticator/GitHubAuthenticator.php
HedicGuibert: I added it because PHPstan was saying the following code was unreachable, but I removed it and it works fine now.

src/Security/Authenticator/GitHubAuthenticator.php
HedicGuibert: Unfortunately, I can't make `$request` nullable because the vendor parent method doesn't define it as nullable.
Regarding the typehint, I added it because PHPStan couldn't find `getFlashBag()` and was throwing an error. I tried to remove it and the error is still there.

lyrixx: I mean, "$request **CAN** be null". That's a fact.
The code **must** handle such case

.circleci/config.yml
```diff
@@ -52,7 +52,7 @@ jobs:
                   when: always
             - run:
                     name: Run PHPStan
-                    command: php -d memory_limit=-1 vendor/bin/phpstan analyse -c ./phpstan.neon.dist src
+                    command: php -d memory_limit=-1 vendor/bin/phpstan analyse -c
```
lyrixx, line 55: `-c` option is useless

phpstan.neon
```diff
@@ -12,5 +13,7 @@ parameters:
 
     symfony:
         container_xml_path: 'var/cache/dev/App_KernelDevDebugContainer.xml'
+        console_application_loader: src/Command/console-application.php
```
lyrixx, line 16: I don't really like theses .php files in the src dir. Could you move them in the `var` dir? thanks

src/Security/Authenticator/GitHubAuthenticator.php
```diff
@@ -116,9 +116,11 @@ public function start(Request $request, AuthenticationException $authException =
 
     public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
     {
-        /** @var Session $session */
-        $session = $this->requestStack->getCurrentRequest()->getSession();
-        $session->getFlashBag()->add('error', 'Your GitHub user is not in the allowed organization, or the membership is private.');
+        if (null !== $request) {
+            /** @var Session $session */
+            $session = $this->requestStack->getCurrentRequest()->getSession();
```
lyrixx: Oh, I did not read the code carefully the first time.
It does not make sens at all. Why using the request stack, when you already have the request? 😨
Wep, the previous code was stupid :/ 

What I was saying, BTW, because you totally miss the point: ` $this->requestStack->getCurrentRequest()` can return `null`
You new code does not make sens either. $request could NOT be null


Anyway: To fix that: use the session from $request. And if PHPStan yield, add an ignore rule in the config file

lyrixx (changes requested):

src/Security/Authenticator/GitHubAuthenticator.php
```diff
@@ -115,7 +115,10 @@ public function start(Request $request, AuthenticationException $authException =
 
     public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
     {
-        $this->requestStack->getCurrentRequest()->getSession()->getFlashBag()->add('error', 'Your GitHub user is not in the allowed organization, or the membership is private.');
+        if (null !== $session = $request->getSession()) {
```
lyrixx, line 118: Why this checks?

phpstan.neon
```diff
@@ -0,0 +1,19 @@
...
+        
+    tmpDir: var/phpstan-tmp
+    inferPrivatePropertyTypeFromConstructor: true
+
+    symfony:
+        container_xml_path: 'var/cache/dev/App_KernelDevDebugContainer.xml'
+        console_application_loader: var/phpstan-objects/console-application.php
+
+    doctrine:
+        objectManagerLoader: var/phpstan-objects/object-manager.php
```
lyrixx, line 19: I don't see theses files in the repo 🤔

src/Security/Authenticator/GitHubAuthenticator.php
```diff
@@ -33,7 +33,7 @@ class GitHubAuthenticator extends SocialAuthenticator
     private $urlGenerator;
     private $allowedGitHubOrganization;
     private $httpClient;
-    private $requestStack;
+    private RequestStack $requestStack;
```
lyrixx, line 36: Dead code

lyrixx:
We'll se later for a full integration with Symfony.

PR #101, opened 2020-12-21
PHPStan bug with console

HedicGuibert:
This is not meant to be merged, this is just to show a bug.

This PR shows a bug with the `console-application.php` file, resulting in this error with the vendors :

PHPStan

I'm not sure whether this comes from us or from PHPStan.

### DIFF
--- a/var/phpstan-utilities/console-application.php
+++ b/var/phpstan-utilities/console-application.php
@@ -1,0 +1,15 @@
+<?php
+
+/*
+ * This file is part of the Starfleet Project.
+ *
+ * (c) Starfleet <msantostefano@jolicode.com>
+ *
+ * For the full copyright and license information,
+ * please view the LICENSE file that was distributed with this source code.
+ */
+
+require __DIR__.'../../../config/bootstrap.php';
+$kernel = new \App\Kernel('phpstan', true);
+
+return new \Symfony\Bundle\FrameworkBundle\Console\Application($kernel);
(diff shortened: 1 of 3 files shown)


HedicGuibert:
So it seems that the original bug is not happening here but we have a new challenger instead

lyrixx:
OK. So we need to disable inlining, but we don't want to do that for dev or prod environement for performance reason. We don't want to do that in test env either, because we want it to be similar to the prod.

Si you'll need to create a new kernel env. Lets call it "phpstan". Then use it.

HedicGuibert:
Fixed and ready to merge !

var/phpstan-utilities/console-application.php
```diff
@@ -0,0 +1,15 @@
...
+ * This file is part of the Starfleet Project.
+ *
+ * (c) Starfleet <msantostefano@jolicode.com>
+ *
+ * For the full copyright and license information,
+ * please view the LICENSE file that was distributed with this source code.
+ */
+
+require __DIR__.'../../../config/bootstrap.php';
+$kernel = new \App\Kernel('phpstan', (bool) $_SERVER['APP_DEBUG']);
```
lyrixx, line 13: No need to ne able to detect the debug mode from $_SERVER. You can hardcode `true`
```suggestion
$kernel = new \App\Kernel('phpstan', true);
```

Issue #69, opened 2020-01-25
Add conferences from tulula


pocky:
URL : https://tulu.la/call-for-papers/
France URL : https://tulu.la/call-for-papers/?in=France

lyrixx:
Done in #104

PR #104, opened 2020-12-24
Add Tulula Fetcher

HedicGuibert:
This PR addresses https://github.com/jolicode/starfleet/issues/69

It adds Tulala to the fetchers list and removes some dead code here and there.
It also handles the cases where the conference is online, because previously it wouldn't display well in the admin (broken image) and the LocationGuesser would set the country to CH for some reason.


src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+    private $tagRepository;
+    private $excludedTags;
+    private $locationGuesser;
+    private $slugger;
+    private HttpClientInterface $client;
+
+    public function __construct(ManagerRegistry $doctrine, LoggerInterface $logger, LocationGuesser $locationGuesser, HttpClientInterface $client)
+    {
+        $this->client = $client;
+        $this->em = $doctrine->getManager();
```
lyrixx, line 63: You inject the MangerRegistry, but you only use it to get the EM. So inject the EM directly

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+        $this->logger = $logger;
+        $this->tagRepository = $this->em->getRepository(Tag::class);
```
lyrixx, line 65: Inject the TagRepository Directly, It will be clearer, and easier to test

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+        $this->excludedTags = $this->em->getRepository(ExcludedTag::class)->findAll();
```
lyrixx, line 66: An object should always be light / fast to instantiate. So not heavy / slow code in the contructor;
1. Inject the ExcludedTag repository
2. call `findAll()` as late as possible

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+        $this->locationGuesser = $locationGuesser;
+        $this->slugger = new AsciiSlugger();
```
lyrixx, line 68: Why no DI?

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+        'sql',
+        'iot',
+    ];
+
+    private $em;
+    private $logger;
+    private $tagRepository;
+    private $excludedTags;
+    private $locationGuesser;
+    private $slugger;
```
lyrixx, line 57: Why no typehint?

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+use Doctrine\Persistence\ManagerRegistry;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\String\Slugger\AsciiSlugger;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class TululaFetcher implements FetcherInterface
+{
+    const SOURCE = 'tulula';
+    const TAGS_SYNONYMS = [
```
lyrixx, line 27: Looks like it could be private?

lyrixx: The naming seems broken. It's not about synonyms, but about filters, isn't?

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+                $cfpEndAt = \DateTimeImmutable::createFromFormat('Y-m-d', $rawConference['cfpDateEnd']);
+                $conference->setCfpEndAt($cfpEndAt);
+            }
+            $conferences[] = $conference;
+        }
+
+        return $conferences;
+    }
+
+    private function queryTululaApi()
```
lyrixx, line 171: 1. why no typehint ?
2. The naming seems bad to me. You don't get an "API", but some events

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+        }
+    }
+
+    public function denormalizeConferences(array $rawConferences, Tag $tag): array
+    {
+        $conferences = [];
+        foreach ($rawConferences as $rawConference) {
+            if (!$rawConference['isOnline']) {
+                // There are cases where Tulula doesn't provide city or state so the string might be empty
+                if (!empty($city = $rawConference['venue']['city'] ?: $rawConference['venue']['state'])) {
```
lyrixx, line 111: This code is not really easy to debug since `empty` is a catch all.
It should use:
```suggestion
                if ($city = $rawConference['venue']['city'] ?? $rawConference['venue']['state'] ?? false) {
```

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+                    $query = sprintf('%s', $city);
```
lyrixx, line 112: seems useless to me

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+            $conference->setName($rawConference['name']);
+            $conference->setStartAt($startDate);
+            $conference->setSiteUrl($rawConference['url']);
+            $conference->addTag($tag);
+
+            if ($rawConference['isOnline']) {
+                $conference->setCity('Online');
+                $conference->setCountry('Online');
+            } else {
+                $conference->setCity($rawConference['venue']['city'] ?: $rawConference['venue']['state'] ?: null);
```
lyrixx, line 140: Why don't you use the $city variable ? (warning it could be not defined)

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+                            city
+                        }
+                        fragment TagData on Tag {
+                            name
+                        }
+                    QUERY
+                ],
+            ]);
+        } catch (TransportExceptionInterface $exception) {
+            $this->logger->error('HttpClient Transport Exception', [
```
lyrixx, line 221: Very nice 👍🏼 
But please add a trailing dot the the sentence (same for all logs or exception)

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+
+            return;
+        }
+
+        if (404 === $response->getStatusCode()) {
+            $this->logger->error('Source URL returns 404', ['url' => $this->getUrl(), 'source' => self::SOURCE]);
+
+            return;
+        }
+
```
lyrixx, line 235: What about all other status code ? like 5XX?
May be it is already handler at another level but I did not see it in the diff

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+        $json = $response->getContent();
+        $body = json_decode($json, true);
```
lyrixx, line 237: there is a $response->toArray() for such things

templates/easy_admin/Conference/country.html.twig
```diff
@@ -1 +1,5 @@
-<img class="country-flag" alt="{{ country_name }} flag" title="{{ country_name }}" src="/bundles/easyadmin/images/flags/{{ value }}.png">
+{% if value|lower == 'online' %}
```
lyrixx, line 1: 🤔 I don't like that because this logic will be spread in all the application.
Please add a method in the class to do that. 
And may be it should be a flag in the class and not a hack with the city / country... Choose the best options here

lyrixx (commented):
Nice start. I did not see the **unit** test class

src/Fetcher/TululaFetcher.php
HedicGuibert: I'm not too sure about this one, maybe we should wait @welcoMattic since he wrote it ?

src/Fetcher/ConfTechFetcher.php
```diff
@@ -49,32 +48,23 @@ class ConfTechFetcher implements FetcherInterface
...
-        $this->serializer = $serializer;
         $this->logger = $logger;
-        $this->conferenceRepository = $this->em->getRepository(Conference::class);
-        $this->tagRepository = $this->em->getRepository(Tag::class);
-        $this->excludedTags = $this->em->getRepository(ExcludedTag::class)->findAll();
+        $this->tagRepository = $tagRepository;
         $this->locationGuesser = $locationGuesser;
-        $this->slugger = new AsciiSlugger();
+        $this->slugger = $slugger;
+        $this->excludedTags = $excludedTagRepository->findAll();
```
lyrixx, line 67: bla bla, constructor, bla bla fast

src/Fetcher/JoindApiFetcher.php
```diff
@@ -15,19 +15,18 @@
...
 use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpClient\HttpClient;
-use Symfony\Component\Serializer\SerializerInterface;
-use Symfony\Component\String\Slugger\AsciiSlugger;
+use Symfony\Component\String\Slugger\SluggerInterface;
 use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
 
-class JoindApiFetcher implements FetcherInterface
+class JoindApiFetcher
```
lyrixx, line 26: same same

src/Fetcher/JoindApiFetcher.php
```diff
@@ -50,28 +49,23 @@ class JoindApiFetcher implements FetcherInterface
...
-        $this->serializer = $serializer;
         $this->logger = $logger;
-        $this->conferenceRepository = $this->em->getRepository(Conference::class);
-        $this->tagRepository = $this->em->getRepository(Tag::class);
-        $this->excludedTags = $this->em->getRepository(ExcludedTag::class)->findAll();
+        $this->tagRepository = $tagRepository;
         $this->locationGuesser = $locationGuesser;
-        $this->slugger = new AsciiSlugger();
+        $this->slugger = $slugger;
+        $this->excludedTags = $excludedTagRepository->findAll();
```
lyrixx, line 68: same same

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+    public function fetch(): \Generator
+    {
+        foreach ($this->tagRepository->findBy(['selected' => true]) as $tag) {
+            $tagKey = array_search($this->slugger->slug(strtolower($tag->getName()))->toString(), self::TAGS_SYNONYMS);
+
+            if (false === $tagKey) {
+                continue;
+            }
+
+            $tagSynonym = self::TAGS_SYNONYMS[$tagKey];
```
lyrixx, line 90: I don't understand this code at all. it's basically the same thing than an \in_array ?

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+use App\Repository\TagRepository;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\String\Slugger\SluggerInterface;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class TululaFetcher implements FetcherInterface
+{
+    const SOURCE = 'tulula';
+    private const TAGS_SYNONYMS = [
```
lyrixx, line 28: TAG_ALLOW_LIST

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+    {
+        return true;
+    }
+
+    public function getUrl(array $params = []): string
+    {
+        return 'https://tulu.la/api/public';
+    }
+
+    public function fetch(): \Generator
```
lyrixx, line 81: The whole method does not make sense.

1. you select all selected tag from the DB
2. then you filter it with some hardcoded item 
3. then for each tag you request the same API endpoint with no parameter

This is a waste of ressource!

1&2 => fetch only tag you want
3 => dont fetch N time data from the API, but only once

The whole method should be re-written from scratch

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,241 @@
...
+        }
+    }
+
+    public function denormalizeConferences(array $rawConferences, Tag $tag): array
+    {
+        $conferences = [];
+        foreach ($rawConferences as $rawConference) {
+            $city = null;
+            if (!$rawConference['isOnline']) {
+                if ($city = $rawConference['venue']['city'] ?: $rawConference['venue']['state']) {
```
lyrixx, line 111: Is it possible to have an falsy value here?

lyrixx (commented):
It's nice to remove code, but now there are too many chances about it.
Please revert all others chances, and you can open a PR later about that
---
I left some comment in the code

src/Fetcher/TululaFetcher.php
HedicGuibert: It is possible to have an empty string, which will result in the city being null.
Do you think we should add a message or handle this case better ?

src/Entity/Tag.php
```diff
@@ -99,6 +99,11 @@ public function setSelected(bool $selected)
         $this->selected = $selected;
     }
 
+    public function getSelected(): bool
```
lyrixx, line 102: IsSelected

src/Entity/Conference.php
```diff
@@ -382,4 +382,13 @@ public function setCity(?string $city): self
 
         return $this;
     }
+
+    public function getOnline(): bool
```
lyrixx, line 386: isOnline

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,238 @@
...
+use App\Repository\ExcludedTagRepository;
+use App\Repository\TagRepository;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\String\Slugger\SluggerInterface;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class TululaFetcher implements FetcherInterface
+{
+    const SOURCE = 'tulula';
```
lyrixx, line 26: Visibility ?

lyrixx (commented):
Much much better

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,235 @@
...
+    private HttpClientInterface $client;
+
+    public function __construct(LoggerInterface $logger, LocationGuesser $locationGuesser, HttpClientInterface $client, TagRepository $tagRepository, ExcludedTagRepository $excludedTagRepository)
+    {
+        $this->client = $client;
+        $this->logger = $logger;
+        $this->tagRepository = $tagRepository;
+        $this->locationGuesser = $locationGuesser;
+        $this->excludedTagsRepository = $excludedTagRepository;
+    }
```
lyrixx, line 64: Can you sort every property in the same order 
1. class property
2. constructor argument
3. constructor assignation
I understand this comment could be boring, but it's really easier to edit arguments later on.

And BTW, the Logger should be the last one (almost always) since it should be optional (almost always)
And to avoid checking if a logger exist before using it, we usually use the Null pattern:
```php
public function __construct(LoggerInterface $logger = null)
{
    $this->logger = $logger ?? new NullLogger()?
}
```

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,235 @@
...
+                    if ($repositoryTag && $repositoryTag->isSelected()) {
+                        $validConferences[$lowerTag][] = $conference;
+                        break;
+                    }
+                }
+            }
+        }
+
+        foreach ($validConferences as $tagName => $conferencesByTag) {
+            yield $this->getUrl() => $this->denormalizeConferences($conferencesByTag, $this->tagRepository->findTagByName($tagName));
```
lyrixx, line 95: Here, and on line 84 you are requesting the same tags over and over.

On this line, this a bit a shame since you already fetch the exact very same tag. Let's store it along with the conference. I would even use a SplObjectStorage here. But looks if it's possible and easy.

**Then** you can add local cache on line 84. You can take inspiration from that:
```php
function getElementById(string $id): ?string
{
    var_dump([__FUNCTION__ => $id]);
    switch ($id) {
        case 'a':
            return 'A';
        case 'b':
            return 'B';
        default:
            return null;
    }
}

$tags = [];
for ($i = 0; $i < 3; $i++) {
    foreach (range('a', 'c') as $id) {
        $tag = array_key_exists($id, $tags) ? $tags[$id] : ($tags[$id] = getElementById($id));
        var_dump([$id => $tag]);
    }
}
```

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,235 @@
...
+    public function getUrl(array $params = []): string
+    {
+        return 'https://tulu.la/api/public';
+    }
+
+    public function fetch(): \Generator
+    {
+        $validConferences = [];
+        foreach ($this->queryTululaEvents() as $conference) {
+            foreach ($conference['tags'] as $tag) {
```
lyrixx, line 80: I don't like the naming here : tag vs lowerTag vs $repositoryTag
Let use:
* $tagName instead of tag & lowerTag
* $tag instead of repositoryTag
I always use this "pattern", it works very well

templates/easy_admin/Conference/country.html.twig
```diff
@@ -1 +1,5 @@
-<img class="country-flag" alt="{{ country_name }} flag" title="{{ country_name }}" src="/bundles/easyadmin/images/flags/{{ value }}.png">
+{% if item.online == true %}
```
lyrixx, line 1:
```suggestion
{% if item.online %}
```

composer.json
```diff
@@ -38,7 +38,7 @@
         "symfony/dotenv": "^5.2",
         "symfony/flex": "^1.0",
         "symfony/framework-bundle": "^5.2",
-        "symfony/http-client": "^5.2",
+        "symfony/http-client": "5.2.*",
```
lyrixx, line 41: Why?

src/Fetcher/TululaFetcher.php
```diff
@@ -0,0 +1,235 @@
...
+    {
+        $validConferences = [];
+        $tags = [];
+
+        foreach ($this->queryTululaEvents() as $conference) {
+            foreach ($conference['tags'] as $tagName) {
+                $tagName = strtolower($tagName['name']);
+
+                if (\in_array($tagName, self::TAG_ALLOW_LIST)) {
+                    if (!\in_array($tagName, $tags)) {
```
lyrixx, line 88: this does not work

tests/Fetcher/TululaFetcherTest.php
```diff
@@ -0,0 +1,203 @@
...
+use App\Repository\ExcludedTagRepository;
+use App\Repository\TagRepository;
+use Prophecy\Argument;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\HttpClient\MockHttpClient;
+use Symfony\Component\HttpClient\Response\MockResponse;
+
+class TululaFetcherTest extends KernelTestCase
+{
+    private function createFetcher(MockResponse $response, ?string $selectedTag = 'php'): TululaFetcher
```
lyrixx, line 28: private method go after public one

tests/Fetcher/TululaFetcherTest.php
```diff
@@ -0,0 +1,203 @@
...
+
+        return $fetcher;
+    }
+
+    /**
+     * @dataProvider provideConferences
+     */
+    public function testFetch(array $fakeConference, ?string $selectedTag = 'php')
+    {
+        $data = json_decode(file_get_contents('tests/Fetcher/fixtures.json'), true);
```
lyrixx, line 65: missing `__DIR__` somewhere:
**always** anchor any path

tests/Fetcher/TululaFetcherTest.php
```diff
@@ -0,0 +1,203 @@
...
+        $data['data']['events']['events'] = $fakeConference;
+
+        $response = new MockResponse(json_encode($data));
```
lyrixx, line 68: Why do json_decode + json_encode?
This is really strange what you did here. You mixed 2 concept : manual fixture and raw fixtures. Use first or second but not both at the same time in the very same test.

Here, you could only use manual fixture.

**BUT** I would like another test with a real raw HTTP response

tests/Fetcher/TululaFetcherTest.php
```diff
@@ -0,0 +1,203 @@
...
+        $result = $this->createFetcher($response, $selectedTag)->fetch();
+
+        // If the tag is not selected, the conference should not be yielded and the response should be empty
+        if (null === $selectedTag) {
+            return self::assertEmpty(iterator_to_array($result));
+        }
+
+        /** @var Conference */
+        $fetchedConference = iterator_to_array($result)[0][0];
```
lyrixx, line 77: Let's make an assertion or 2 before calling `[0]`, like assertCount()

tests/Fetcher/TululaFetcherTest.php
```diff
@@ -0,0 +1,203 @@
...
+        self::assertSame($fakeConference[0]['name'], $fetchedConference->getName());
+        self::assertSame($fakeConference[0]['url'], $fetchedConference->getSiteUrl());
+        $start = $fetchedConference->getStartAt();
+        $end = $fetchedConference->getEndAt();
+        $cfpEnd = $fetchedConference->getCfpEndAt();
+        self::assertTrue($start instanceof \DateTimeImmutable);
+        self::assertTrue($end instanceof \DateTimeImmutable);
+        self::assertTrue($cfpEnd instanceof \DateTimeImmutable);
+        self::assertGreaterThan($start, $end);
+        self::assertGreaterThan($cfpEnd, $start);
```
lyrixx, line 88: It looks like you are testing nothing here

tests/Fetcher/TululaFetcherTest.php
```diff
@@ -0,0 +1,203 @@
...
+        self::assertSame($fakeConference[0]['cfpUrl'], $fetchedConference->getCfpUrl());
+        self::assertSame($fakeConference[0]['isOnline'], $fetchedConference->isOnline());
+        self::assertSame($fakeConference[0]['slug'], $fetchedConference->getSlug());
+
+        if ($fakeConference[0]['venue']) {
+            self::assertSame($fakeConference[0]['venue']['countryCode'], $fetchedConference->getCountry());
+            self::assertNotEmpty($fetchedConference->getCity());
+            self::assertSame(
+                $fakeConference[0]['venue']['city'] ?: $fakeConference[0]['venue']['state'],
```
lyrixx, line 97: Here you duplicate the logic from the tested code. you should not

lyrixx (commented):
It's a good start 👍🏼

tests/Fetcher/TululaFetcherTest.php
```diff
@@ -0,0 +1,225 @@
...
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\HttpClient\MockHttpClient;
+use Symfony\Component\HttpClient\Response\MockResponse;
+
+class TululaFetcherTest extends KernelTestCase
+{
+    /**
+     * @dataProvider provideConferences
+     */
+    public function testFetch(array $fakeConference, ?string $selectedTag = 'php')
```
lyrixx, line 31:
```suggestion
    public function testFetch(array $rawConferences, ?string $selectedTag = 'php')
```

tests/Fetcher/TululaFetcherTest.php
```diff
@@ -0,0 +1,225 @@
...
+
+        // If the tag is not selected, the conference should not be yielded and the response should be empty
+        if (null === $selectedTag) {
+            return self::assertEmpty(iterator_to_array($result));
+        }
+
+        $fetchedConferences = iterator_to_array($result)[0];
+        self::assertCount(\count($fakeConference), $fetchedConferences);
+
+        for ($i = 0; $i < \count($fetchedConferences); ++$i) {
```
lyrixx, line 46: Why don't you use a foreach?

tests/Fetcher/TululaFetcherTest.php
```diff
@@ -0,0 +1,225 @@
...
+            /** @var Conference */
+            $fetchedConference = $fetchedConferences[$i];
+
+            self::assertSame($fakeConference[$i]['name'], $fetchedConference->getName());
+            self::assertSame($fakeConference[$i]['url'], $fetchedConference->getSiteUrl());
+            self::assertTrue($fetchedConference->getStartAt() instanceof \DateTimeImmutable);
+            self::assertTrue($fetchedConference->getEndAt() instanceof \DateTimeImmutable);
+            self::assertTrue($fetchedConference->getCfpEndAt() instanceof \DateTimeImmutable);
```
lyrixx, line 54: assertInstanceOf, and I would add an assert on the value too

tests/Fetcher/TululaFetcherTest.php
```diff
@@ -0,0 +1,225 @@
...
+            if ($fakeConference[$i]['venue']) {
+                self::assertSame($fakeConference[$i]['venue']['countryCode'], $fetchedConference->getCountry());
+
+                if ($fakeConference[$i]['venue']['city']) {
+                    self::assertSame($fakeConference[$i]['venue']['city'], $fetchedConference->getCity());
+                } elseif ($fakeConference[$i]['venue']['state']) {
+                    self::assertSame($fakeConference[$i]['venue']['state'], $fetchedConference->getCity());
+                } else {
+                    self::assertEmpty($fetchedConference->getCity());
+                }
```
lyrixx, line 68: You duplicate the logic you are currently testing. you must not.

tests/Fetcher/TululaFetcherTest.php
```diff
@@ -0,0 +1,225 @@
...
+            self::assertSame($fakeConference[$i]['name'], $fetchedConference->getName());
+            self::assertSame($fakeConference[$i]['url'], $fetchedConference->getSiteUrl());
+            self::assertTrue($fetchedConference->getStartAt() instanceof \DateTimeImmutable);
+            self::assertTrue($fetchedConference->getEndAt() instanceof \DateTimeImmutable);
+            self::assertTrue($fetchedConference->getCfpEndAt() instanceof \DateTimeImmutable);
+            self::assertSame($fakeConference[$i]['cfpUrl'], $fetchedConference->getCfpUrl());
+            self::assertSame($fakeConference[$i]['isOnline'], $fetchedConference->isOnline());
+            self::assertSame($fakeConference[$i]['slug'], $fetchedConference->getSlug());
+
+            if ($fakeConference[$i]['venue']) {
```
lyrixx, line 59: You duplicate the logic you are currently testing. you must not.

lyrixx (commented):
This is better :) 

But I **still** miss a test with a raw response

Some parts of the code are not tested. But it's not really problematic, since I think we'll update the overall design of how the fetcher works (Tag filtering should not be inside the fetching)

lyrixx:
Good job. Thanks

PR #105, opened 2020-12-28
Add phpstan task

HedicGuibert:
This adds an invoke task to run PHPStan

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -154,6 +154,15 @@ def tests(c):
 
 
 @task
+def phpstan(c):
+    """
+    Runs PHPStan
+    """
+    with Builder(c):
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -153,6 +153,15 @@ def tests(c):
         docker_compose_run(c, 'php ./vendor/bin/simple-phpunit')
 
 
+@task
+def phpstan(c):
+    """
+    Runs PHPStan
+    """
+    with Builder(c):
+        docker_compose_run(c, 'php ./vendor/bin/phpstan analyse')
```
lyrixx, line 162: It should be:

```suggestion
        docker_compose_run(c, 'vendor/bin/phpstan analyse')
```

But I see all the file is like that :( so let's keep it as it

lyrixx (approved):

Issue #106, opened 2020-12-29
Properly handle online conference


lyrixx:
ATM, we are using city="online" to do that. This is not really clean.

* [x] Let's add a real flag in the DB. 
* [x] Existing data should be migrated too
* [x] Update all part of the code using this

PR #109, opened 2021-01-11
Online feature

HedicGuibert:
Addresses https://github.com/jolicode/starfleet/issues/106

It adds an Online icon rather than just displaying 'Online' for online conferences. It also adds a real online field in the database rather than trying to know if the conference is online by checking its city field.

It also fixes a bug I introduced earlier in the show view where the page would break because I forgot to change a variable name (oupsi !).

However, it requires https://github.com/jolicode/starfleet/pull/104 to work, so this should not be merged until this one is merged


lyrixx:
Could you rebase? Thanks

migrations/Version20210111100854.php
```diff
@@ -0,0 +1,33 @@
...
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20210111100854 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE conference ADD online BOOLEAN DEFAULT NULL');
+        $this->addSql('UPDATE conference SET online = false');
+        $this->addSql('UPDATE conference SET online = true WHERE LOWER(city) = \'online\'');
```
lyrixx, line 25: Well done 👍🏼

src/DataFixtures/FixtureBuilder.php
```diff
@@ -119,19 +119,27 @@ public static function createConference(array $description = []): Conference
...
         $conference->setCfpEndAt($description['cfpEndAt']);
+        $conference->setOnline($description['online']);
+
+        if (!$description['online']) {
+            $conference->setCity($description['city']);
+            $conference->setCountry($description['country']);
+        } else {
+            $conference->setCity('Online');
+            $conference->setCountry(null);
+        }
```
lyrixx, line 142: Let's use the positive form before the negative on. I'll be easier to read:



```suggestion
        if ($description['online']) {
            $conference->setCity('Online');
            $conference->setCountry(null);
        } else {
            $conference->setCity($description['city']);
            $conference->setCountry($description['country']);
        }
```

BTW, the following line is useless, isn't?
```php
            $conference->setCountry(null);
```

src/Entity/Conference.php
```diff
@@ -127,6 +127,11 @@ class Conference
      */
     private ?string $city = null;
 
+    /**
+     * @ORM\Column(type="boolean", nullable=true)
+     */
+    private ?bool $online = false;
```
lyrixx, line 133: No need for `?bool` since you initialize the value + It does not make sens to set null where.
Same in the setter.

src/Fetcher/TululaFetcher.php
```diff
@@ -149,6 +149,7 @@ public function denormalizeConferences(array $rawConferences, Tag $tag): array
             if ($rawConference['isOnline']) {
                 $conference->setCity('Online');
                 $conference->setCountry(null);
```
lyrixx, line 151: Not useful anymore?

lyrixx (commented):
Very good 👍🏼 
I let few minor comment. Once fixed + green, you can merge yourself :)

templates/conferences/show.html.twig
```diff
@@ -49,7 +49,11 @@
                             <i class="fa fa-map"></i>&nbsp;
                             {{ 'conferences.show.sidebar.location_title'|trans }}
                         </p>
-                        <p>{{ conference.location }}</p>
+                        {% if conference.online %}
+                            <p>Online</p>
```
welcoMattic, line 53: Missing translation

HedicGuibert: You want me to translate 'Online' ? I guess I'm missing something here because to me it looks like it is common to all languages ?
Or maybe I'm wrong and this is standard only to latin languages

welcoMattic: No matter if it translates the same in all languages or not, we must centralize all UI texts in translations files (to make them reusable and updatable all at once).

Issue #108, opened 2021-01-11
Extract filtering / tag grouping from fetch. It not the right place to do that


lyrixx:
The fetch has too much responsability. 

1. The public method in the interface that are not used outside must be removed from the interface
2. The tag filter + group should be move outside fetchers

PR #110, opened 2021-01-12
Rename ExcludedTag and move filtering from fetchers

HedicGuibert:
This renames the ExcludedTag entity to ConferenceFilter because it was confusing.
It also removes the filtering from the fetcher and puts it in the fetch command.

Refers https://github.com/jolicode/starfleet/issues/108


src/Command/FetchConferencesCommand.php
```diff
@@ -197,4 +197,18 @@ protected function updateExistingConference(Conference $existingConference, Conf
 
         return $updated;
     }
+
+    private function shouldBeIgnored(Conference $conference): bool
+    {
+        foreach ($this->conferenceFilterRepository->findAll() as $conferenceFilter) {
```
welcoMattic, line 203: This `findAll` call will be repeated many time, as the `shouldBeIgnored` function is called inside 2 foreach loops. May be an initial call that store the result in a variable will be faster and avoir many DB queries?

src/Command/FetchConferencesCommand.php
```diff
@@ -68,9 +68,9 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                 continue;
             }
 
-            foreach ($fetcher->fetch() as $url => $conferences) {
+            foreach ($fetcher->fetch() as $conferences) {
                 foreach ($conferences as $conference) {
```
lyrixx, line 72: Let's get ride of this double foreach

src/Command/FetchConferencesCommand.php
```diff
@@ -197,4 +197,18 @@ protected function updateExistingConference(Conference $existingConference, Conf
...
+    private function shouldBeIgnored(Conference $conference): bool
+    {
+        foreach ($this->conferenceFilterRepository->findAll() as $conferenceFilter) {
+            if (fnmatch($conferenceFilter->getName(), $conference->getName(), FNM_CASEFOLD)) {
+                $conference->setExcluded(true);
+
+                return true;
+            }
+        }
+        $conference->setExcluded(false);
```
lyrixx, line 210: Lets move excluded = false into the conference constructor

lyrixx (commented):
Nice 👍🏼 
3 minors things to update and it'll be OK

src/Command/FetchConferencesCommand.php
HedicGuibert: Actually excluded is false by default so I just removed this line

lyrixx (approved):
Very good 👏🏼  👏🏼  👏🏼

Issue #112, opened 2021-01-13
Update the concept of "Tag" -> FetcherConfiguration


lyrixx:
Tag does not really work ATM. It's used to filter what we want to grab thanks to fetchers.
But fetchers has also their own "tag" that is own by their API.

So the goal of this issue is:

* Remove the Tag concept (entity + all related code)
  * for the entity Conference, use a plain array in place of the current relation (please, migrate all existing data)
* Add a new concept : FetcherConfiguration
  * add a new entity: `id`, `className`, `configuration`
  * add `public function buildForm(): FormBuilderInterface` (not sure ATM about the return type) in `FetcherInterface` 
  * use the form in the admin, to configure each fetcher
  * **If it's easy**, let's be nice: add a way to create a new record for each fetcher if it does not exists in the DB yet
  * At runtime, each fetcher should get it's configuration. But IMHO, this should be delegate to the service that call the fetcher. So maybe update the interface `FetcherInterface::fetch(array $configuration): \Generator`

PR #114, opened 2021-01-18
Fetcher config

HedicGuibert:
Addresses https://github.com/jolicode/starfleet/issues/112

This is for now a WIP.

This refactors fetchers to no longer use a Tag entity but instead a configuration entity that will set the way the user want to fetch its data. 

Now, the fetchers holds all the tags available from the sources (or at least the ones I could find). It will be used in EasyAdmin to select the ones the user is interrested in. The checked ones will be sent to the config, which will be used to filter the command results.


src/Entity/FetcherConfiguration.php
```diff
@@ -0,0 +1,70 @@
...
+     */
+    private string $fetcherName;
+
+    /**
+     * This is an array that will configure the way events are fetched.
+     * All fetchers are different so there is no common configuration. However, all have an `active` key which will disable the fetcher if set to false.
+     *
+     * @ORM\Column(type="array")
+     */
+    private array $configuration = ['active' => true];
```
lyrixx, line 40: Active could have a dedicated entry in the DB.

src/Command/FetchConferencesCommand.php
```diff
@@ -67,11 +73,19 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         foreach ($this->fetchers as $fetcher) {
             $fetchersSection->overwrite(sprintf('Processing %d/%d fetchers : %s', ++$currentFetcher, $nbFetchers, \get_class($fetcher)));
 
-            if (!$fetcher->isActive()) {
-                continue;
+            /** @var FetcherConfiguration */
```
lyrixx, line 76: This is already supported by `@method` in the repo

src/Fetcher/ConfTechFetcher.php
```diff
@@ -13,117 +13,90 @@
...
         foreach ([date('Y'), date('Y', strtotime('+1 year'))] as $date) {
-            foreach ($this->tagRepository->findBy(['selected' => true]) as $tag) {
-                $tagKey = array_search($this->slugger->slug(strtolower($tag->getName()))->toString(), self::TAGS_SYNONYMS);
+            if (0 === \count($configuration) || 0 === \count($configuration['tags'])) {
+                $this->logger->notice('This fetcher is not configured and fetches all data from the source. If this behaviour is not intended, you should add a configuration in the admin');
 
-                if (false === $tagKey) {
-                    continue;
-                }
+                foreach (self::SOURCE_AVAILABLE_TAGS as $tag) {
```
lyrixx, line 78: Let's do nothing if no tags are configured
Same same for others fetchers

src/Fetcher/ConfTechFetcher.php
```diff
@@ -176,4 +152,31 @@ public function denormalizeConference(array $rawConference, Tag $tag): ?Conferen
...
+            $this->logger->error(sprintf('Source URL returns %d', $statusCode), [
+                'url' => $url,
+                'source' => self::SOURCE,
+            ]);
+
+            return null;
+        }
+        $result = $response->toArray();
+
+        return $result;
```
lyrixx, line 180:
```suggestion

        return $response->toArray();
```

src/Fetcher/JoindApiFetcher.php
```diff
@@ -13,118 +13,125 @@
...
+            foreach ($this->queryJoindIn() as $conference) {
+                yield $this->denormalizeConference($conference);
             }
 
-            $data = $response->toArray();
+            return;
+        }
 
-            if (0 === $data['meta']['total']) {
+        foreach ($this->queryJoindIn() as $conference) {
```
lyrixx, line 114: Can we give joind the tags we want? pre filtering is better than post filtering

src/Repository/ConferenceRepository.php
```diff
@@ -53,9 +53,9 @@ public function findAttendedConferences(): array
     public function findAttendedConferencesByTag(string $tag): array
     {
         return $this->createAttendedQueryBuilder()
-            ->leftJoin('c.tags', 't')
-            ->andWhere('t.name = :tagName')
-            ->setParameter('tagName', $tag)
+            ->select('c')
+            ->andWhere('c.tags LIKE :tagName')
+            ->setParameter('tagName', '%"'.$tag.'"%')
```
lyrixx, line 58: 😬 Let's use a real SQL function for that. something like `JSON_CONTAINS()`

There are some doctrine extension OSS. Copy what you need

src/Entity/Conference.php
```diff
@@ -399,4 +376,16 @@ public function isOnline(): bool
...
+    {
+        $this->tags[] = $tag;
+
+        return $this;
+    }
+
+    public function getTags(): array
+    {
+        return $this->tags;
+    }
```
welcoMattic, line 390: Is not a `removeTag` method missing here?

src/Fetcher/ConfTechFetcher.php
```diff
@@ -13,117 +13,90 @@
...
+        $this->logger = $logger ?: new NullLogger();
     }
 
-    public function fetch(): \Generator
+    public function fetch(array $configuration = []): ?\Generator
     {
         foreach ([date('Y'), date('Y', strtotime('+1 year'))] as $date) {
-            foreach ($this->tagRepository->findBy(['selected' => true]) as $tag) {
-                $tagKey = array_search($this->slugger->slug(strtolower($tag->getName()))->toString(), self::TAGS_SYNONYMS);
+            if (0 === \count($configuration) || 0 === \count($configuration['tags'])) {
```
welcoMattic, line 75: Shouldn't we pass the FetcherConfiguration object here (which will act as a Value Object). May be with a dedicated method in the FetcherConfiguration to check if it's ok will be relevant? Something like `$fetcherConfig->isComplete()`?

src/Fetcher/ConfTechFetcher.php
```diff
@@ -13,117 +13,90 @@
...
-                        'source' => self::SOURCE,
-                    ]);
-                    continue;
-                }
+                return;
+            }
 
-                $data = $response->toArray();
+            foreach ($configuration['tags'] as $tag) {
+                $url = sprintf('https://github.com/tech-conferences/conference-data/tree/main/conferences/%s/%s.json', $date, $tag);
```
welcoMattic, line 90: This url should be stored in a private constant, no?

src/Repository/ConferenceRepository.php
welcoMattic: If I remember well, there's already the needed extension. I has to use it for the Workflow part, to retrieve if an object had one of the given states (stored in JSON)

src/Entity/Conference.php
lyrixx: Conference are readonly IIRC. So let's not add not used code

src/Fetcher/ConfTechFetcher.php
lyrixx: Not sure it's useful. **and** it's not possible to add custom method on $fetcher config since it's generic for all fetcher

src/Fetcher/ConfTechFetcher.php
lyrixx: Up to @HedicGuibert ... To me: there are no real values to create a constant for that... but not a big deal.

Anyway, I want to @HedicGuibert rewrite this fetcher with a `git clone`

src/Entity/Conference.php
welcoMattic: Oh yes, you're right. My bad

HedicGuibert:
I pushed a first version of the admin. Some things are still missing but here it what it looks like now: 

List view :
list

Edit view:
edit

The select:
select2

src/Repository/FetcherConfigurationRepository.php
```diff
@@ -0,0 +1,44 @@
...
+
+    public function findOneOrCreate(string $name): FetcherConfiguration
+    {
+        $config = $this->findOneBy(['fetcherName' => $name]);
+
+        if (null === $config) {
+            $config = new FetcherConfiguration();
+            $config->setFetcherName($name);
+            $this->_em->persist($config);
+            $this->_em->flush();
```
lyrixx, line 39: Do not persist + flush here.

Usually we should never flush outside an entry point (controller, command, message handler). => actually follow **always** this advice

Flush automatically save everything that is in the UOW (Unit Of Work). So here one may think that only the fetcher will be saved, but depending on the previous executed code, many other things could be save. This is a source of bug and hard time to debug.

templates/fetchers/edit.html.twig
```diff
@@ -0,0 +1,65 @@
...
+        </legend>
+        <div class="row">
+            <div class="col-12">
+                {{ form_row(form.tags) }}
+            </div>
+            {% if form.allowEmptyTags is defined %}
+                <div class="col-12">
+                    {{ form_row(form.allowEmptyTags) }}
+                </div>
+            {% endif %}
```
lyrixx, line 34: Nope :/ 
Nothing spefic is allowed here.
IMHO, the best way is to call form_form(form.configuration).

And maybe we will need to nest the form (as said few days ago).
It'll be cleaner and avoir the "hack" in the admin form `unset($data['active])`.

lyrixx (commented):
This is very nice 👍🏼

src/Controller/Admin/FetcherController.php
```diff
@@ -0,0 +1,115 @@
...
+        $class = sprintf('App\Fetcher\%s', $fetcher);
+
+        if (!$this->serviceProvider->has($class)) {
+            throw $this->createNotFoundException("Required fetcher $fetcher not found");
+        }
+
+        $fetcherConfig = $this->fetcherConfigurationRepository->findOneOrCreate($fetcher);
+        $fetcher = $this->serviceProvider->get($class);
+
+        $formFactory = $this->container->get('form.factory');
```
welcoMattic, line 72: Isn't it possible to inject this in the constructor?

src/Controller/Admin/FetcherController.php
```diff
@@ -0,0 +1,115 @@
...
+                'label' => 'Active',
+                'required' => false,
+            ])
+            ->add($fetcher->buildForm($fetcherConfig, $formFactory))
+            ->getForm();
+
+        if ($form->handleRequest($request)->isSubmitted() && $form->isValid()) {
+            $this->em->persist($fetcherConfig);
+            $this->em->flush();
+            $this->addFlash('success', 'Fetcher successfuly edited');
```
welcoMattic, line 85:
```suggestion
            $this->addFlash('success', 'Fetcher successfully edited');
```

welcoMattic: I think it would be better with a translation key under the `admin` domain

src/Fetcher/ConfTechFetcher.php
welcoMattic: Oh, with a GitHub API Client? It's a very nice idea :+1:

lyrixx (approved):
LGTM 👍🏼 

Could you rebase? If green, You can merge yourself :)

src/Fetcher/.gitignore
```diff
@@ -0,0 +1 @@
+conftech_data/
```
lyrixx, line 1: Nope, use the `var/tmp` folders instead. It's done for that :)

src/Fetcher/ConfTechFetcher.php
```diff
@@ -13,147 +13,148 @@
...
-                if (false === $tagKey) {
-                    continue;
-                }
+        if ($this->filesystem->exists($conftechFile = __DIR__.'/conftech_data/conferences')) {
+            $process = new Process(['git', 'pull'], $conftechFile);
+        } else {
+            $process = new Process(['git', 'clone', '--depth', '1', 'https://github.com/tech-conferences/conference-data/', 'conftech_data/'], __DIR__);
+        }
+        $process->mustRun();
+        $process->getOutput();
```
lyrixx, line 91: dead code

src/Fetcher/ConfTechFetcher.php
```diff
@@ -13,147 +13,148 @@
...
 
-                if (404 === $response->getStatusCode()) {
-                    $this->logger->error('Source URL returns 404', [
-                        'url' => $url,
-                        'source' => self::SOURCE,
-                    ]);
+                if ($this->filesystem->exists($path)) {
+                    $conferences = json_decode(file_get_contents($path), true);
+                } else {
                     continue;
```
lyrixx, line 100: Inverse the condition and bailout asap

lyrixx (approved):

lyrixx:
Well done 👍🏼 👏🏼

PR #117, opened 2021-01-28
Update code to PHP7.4 features

HedicGuibert:
This adds some missing TypeHint to the code.


src/Controller/SecurityController.php
```diff
@@ -58,22 +58,23 @@ public function connectGoogleCheckAction(Request $request, ClientRegistry $clien
...
-    public function connectGitHubCheckAction(Request $request)
-    {
-    }
+    // Looks like a TODO ? It is never called
+    // /**
+    //  * @Route("/connect/github/check", name="connect_github_check")
+    //  */
+    // public function connectGitHubCheckAction(Request $request)
+    // {
+    // }
```
welcoMattic, line 72: No, it's used by the GitHubAuthenticationProvider

welcoMattic (dismissed):

HedicGuibert:
I ended up deleting the assignment of a unique ID with `uuid_create` in the fixtures because it wasn't useful (and it was generating a string where we wanted an int).
As a result, I also removed the ReflectionHelper, since it was only used to assign IDs generated by `uuid_create`

src/Entity/Conference.php
```diff
@@ -11,6 +11,7 @@
 
 namespace App\Entity;
 
+use DateTimeInterface;
```
lyrixx, line 14: We don't do that

src/Fetcher/ConfTechFetcher.php
```diff
@@ -72,6 +72,11 @@ public function __construct(LocationGuesser $locationGuesser, Filesystem $filesy
         $this->logger = $logger ?: new NullLogger();
     }
 
+    /**
+     * @param array<mixed> $configuration
+     *
+     * @return \Generator<Conference>\null
```
lyrixx, line 78: the `\null` seems odd to me

src/Fetcher/FetcherInterface.php
```diff
@@ -11,10 +11,16 @@
...
 
+use App\Entity\Conference;
 use Symfony\Component\Form\FormBuilderInterface;
 
 interface FetcherInterface
 {
+    /**
+     * @param array<mixed> $configuration
+     *
+     * @return \Generator<Conference>\null
```
lyrixx, line 22: Same error!

I usually don't repeat the same comment on each error because

1. It's boring for me and you
2. It takes too much place in the UI
3. I expect you to learn, and try to enhance your code according to your new learning, and not fixing all my comments without thinking. The goal is not to be the fastest, but the smarter :)

so from now on, please fix all occurrence of an error :)

src/Doctrine/DBAL/Types/Jsonb.php
```diff
@@ -49,6 +50,8 @@ public function convertToDatabaseValue($value, AbstractPlatform $platform): ?str
      * Converts a value from its database representation to its PHP representation of this type.
      *
      * @param string|null $value the value to convert
+     *
+     * @return array<mixed>
```
lyrixx, line 54: Could be null actually

src/Fetcher/JoindApiFetcher.php
```diff
@@ -203,6 +209,7 @@ public function denormalizeConference(array $rawConference): ?Conference
         return $conference;
     }
 
+    /** @return array<array>\null */
```
lyrixx, line 212: ...

src/Fetcher/TululaFetcher.php
```diff
@@ -210,6 +216,7 @@ private function denormalizeConference(array $rawConference): ?Conference
         return $conference;
     }
 
+    /** @return array<array>\null */
```
lyrixx, line 219: Again

src/Fetcher/JoindApiFetcher.php
HedicGuibert: I left these on purpose because we were returning null explicitely in the method, but we could return an empty array (which I considered but didn't really like)

lyrixx:
Could you rebase? Thanks

lyrixx (approved):
Thanks 👍🏼

Issue #93, opened 2020-12-15
Foreign key violation when deleting a user or a conference


HedicGuibert:
There is a bug when trying to delete a user or a conference in Easy Admin.

If you try to delete a user, the foreign key of that user is still referenced in the participation table, raising this exception :
![User Exception](https://user-images.githubusercontent.com/71645693/102241059-d9186c00-3ef8-11eb-9ec7-a9feba97eea9.png)

The same occurs with some conferences, but not all of them (I don't know why, it may be because of the fixtures), because the key is still referenced in the submit table.

A quick solution would be to remove the possibility to delete a user or a conference in the admin, but ultimately a decision should should be taken about whether or not we want to completely remove all participations of deleted users. Same goes with conferences : if we delete a conference, do we want to delete all submits referencing this conference ?

welcoMattic:
Good catch! May be it's time to implement soft delete or archive for users and conferences? It could be relevant because the organization could want to keep the history of all users and and therefore the associated conferences.

lyrixx:
@welcoMattic It's your project, you have to take a decision :) 
Personally, If we could avoid soft delete, I would! (GDRP at least)
Instead I would "SET NULL" where needed. And of course, the frontend would be update everywhere we display a username

welcoMattic:
I'm not sure that we should worry about GDPR as Starfleet can be installed by everyone. But, ok let's assume it's better to provide à GDPR compliant software ;)
Ok to "SET NULL" where needed and adapt the frontend :+1:

lyrixx:
The more I think about it, the less I understand it. IMHO this does not make sense to remove a conference when people attend to it or when a submission has been made. I would forbid such action. 

@welcoMattic wdyt ?

welcoMattic:
Indeed, I agree to forbid the deletion of conferences. As most of the time they are automatically created by fetchers, and not manually. Let's forbid this action, and if it's annoying in the future, we will work back on it.

PR #118, opened 2021-01-29
Fix user and conference removal

HedicGuibert:
Refs https://github.com/jolicode/starfleet/issues/93

Set `null` on related entities when removing a user or a conference

### DIFF
--- a/src/Controller/Admin/AdminController.php
+++ b/src/Controller/Admin/AdminController.php
@@ -17,7 +17,6 @@ use App\Entity\Participation;
 use App\Entity\Submit;
 use App\Entity\Talk;
 use App\Enum\Workflow\Transition\Participation as ParticipationTransition;
-use Psr\Log\LoggerInterface;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
 use Symfony\Component\Form\Form;
 use Symfony\Component\HttpFoundation\Response;
@@ -26,13 +25,11 @@ use Symfony\Component\Workflow\Registry as WorkflowRegistry;
 
 class AdminController extends EasyAdminController
 {
-    protected $workflowRegistry;
-    protected $logger;
+    protected WorkflowRegistry $workflowRegistry;
 
-    public function __construct(WorkflowRegistry $workflowRegistry, LoggerInterface $logger)
+    public function __construct(WorkflowRegistry $workflowRegistry)
     {
         $this->workflowRegistry = $workflowRegistry;
-        $this->logger = $logger;
     }
 
     public function excludeConferenceAction(): Response
--- a/config/packages/easy_admin/conference.yaml
+++ b/config/packages/easy_admin/conference.yaml
@@ -2,7 +2,7 @@ easy_admin:
     entities:
         PassedConference:
             class: App\Entity\Conference
-            disabled_actions: ['edit']
+            disabled_actions: ['edit', 'delete']
             list:
                 filters: ['name', 'city', 'country', 'startAt', 'tags', 'source']
                 actions:
@@ -17,7 +17,7 @@ easy_admin:
                     - { property: 'tags', type: 'array' }
         NextConference:
             class: App\Entity\Conference
-            disabled_actions: ['show']
+            disabled_actions: ['show', 'delete']
             list:
                 filters: ['name', 'city', 'country', 'startAt', 'tags', 'source']
                 actions:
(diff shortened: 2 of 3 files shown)


src/Controller/Admin/AdminController.php
```diff
@@ -225,4 +233,70 @@ protected function persistTalkEntity(Talk $talk, Form $newForm)
...
+        }
+
+        $participations = $this->participationRepository->findBy(['conference' => $conference]);
+
+        foreach ($participations as $participation) {
+            $participation->setConference(null);
+        }
+
+        $this->em->remove($conference);
+        $this->em->flush();
```
welcoMattic, line 300: You must centralize the duplicated code in a private method then call it in the 3 remove*ConferencesEntity methods

welcoMattic (changes requested):

welcoMattic (approved):

src/Controller/Admin/AdminController.php
```diff
@@ -26,13 +30,17 @@
 
 class AdminController extends EasyAdminController
 {
-    protected $workflowRegistry;
-    protected $logger;
+    protected WorkflowRegistry $workflowRegistry;
+    protected ParticipationRepository $participationRepository;
+    protected SubmitRepository $submitRepository;
+    protected ?LoggerInterface $logger;
```
lyrixx, line 36: There is always a logger defined

src/Controller/Admin/AdminController.php
```diff
@@ -225,4 +233,49 @@ protected function persistTalkEntity(Talk $talk, Form $newForm)
...
+        }
+
+        $participations = $this->participationRepository->findBy(['conference' => $conference]);
+
+        foreach ($participations as $participation) {
+            $participation->setConference(null);
+        }
+
+        $this->em->remove($conference);
+        $this->em->flush();
```
lyrixx, line 279: If you configure property the DB (and doctrine), The DB will do it for you.
Here, you load tons of data ; you should not. The DB must handle that

migrations/Version20210208110825.php
```diff
@@ -0,0 +1,34 @@
...
+ * For the full copyright and license information,
+ * please view the LICENSE file that was distributed with this source code.
+ */
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20210208110825 extends AbstractMigration
```
lyrixx, line 19: Let's merge this migration with the previous one

src/Controller/Admin/AdminController.php
```diff
@@ -26,13 +29,17 @@
...
+    protected LoggerInterface $logger;
 
-    public function __construct(WorkflowRegistry $workflowRegistry, LoggerInterface $logger)
+    public function __construct(WorkflowRegistry $workflowRegistry, ParticipationRepository $participationRepository, SubmitRepository $submitRepository, ?LoggerInterface $logger = null)
     {
         $this->workflowRegistry = $workflowRegistry;
-        $this->logger = $logger;
+        $this->participationRepository = $participationRepository;
+        $this->submitRepository = $submitRepository;
+        $this->logger = $logger ?: new NullLogger();
```
lyrixx, line 42: Looks like this can be revert

src/Entity/Participation.php
```diff
@@ -27,15 +27,15 @@ class Participation
 
     /**
      * @ORM\ManyToOne(targetEntity="App\Entity\Conference", inversedBy="participations", cascade={"persist"})
-     * @ORM\JoinColumn(nullable=false)
+     * @ORM\JoinColumn(nullable=true)
      */
-    private $conference;
+    private ?Conference $conference;
```
lyrixx, line 32: It does not make sens to me. Why would we want to remove a Conference where someone attended?

lyrixx (commented):
Acutally, I don't understand the goal of this PR.

Why do we want to support conference removal?
And user removal?

HedicGuibert:
This addresses https://github.com/jolicode/starfleet/issues/93 where we decided to set null when a user or a conference is deleted

HedicGuibert:
So as mentioned in https://github.com/jolicode/starfleet/issues/93 I ended up reverting everything and only disabling the possibility to delete a user or a conference

lyrixx (approved):
Thanks

Issue #38, opened 2019-07-05
Improve admin panel with Role based access


welcoMattic:
As admin, I can manage User CRUD, Conference CRUD, Participation CRUD (+ workflow), Tag CRUD.

As user, I can only manage my Talks CRUD, my Participation CRUD.

As it is said here https://github.com/EasyCorp/EasyAdminBundle/pull/2810 we should be able to use native permission feature of EasyAdmin soon :tada:

lyrixx:
Could it be closed thanks to #120?

PR #120, opened 2021-02-10
Add voter for users

HedicGuibert:
This references https://github.com/jolicode/starfleet/issues/38 and allows users to only update (and even see) the talks and permissions that they own

### DIFF
--- a/config/packages/easy_admin/participation.yaml
+++ b/config/packages/easy_admin/participation.yaml
@@ -7,6 +7,7 @@ easy_admin:
                 list: 'easy_admin/Participation/list.html.twig'
             list:
                 filters: ['conference', 'participant']
+                item_permission: ['PARTICIPATION_SHOW']
                 fields:
                     - { property: 'conference' }
                     - { property: 'participant' }
--- a/src/Security/Voter/ParticipationVoter.php
+++ b/src/Security/Voter/ParticipationVoter.php
@@ -1,0 +1,57 @@
+<?php
+
+/*
+ * This file is part of the Starfleet Project.
+ *
+ * (c) Starfleet <msantostefano@jolicode.com>
+ *
+ * For the full copyright and license information,
+ * please view the LICENSE file that was distributed with this source code.
+ */
+
+namespace App\Security\Voter;
+
+use App\Entity\Participation;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Authorization\Voter\Voter;
+use Symfony\Component\Security\Core\Role\RoleHierarchyInterface;
+use Symfony\Component\Security\Core\User\UserInterface;
+
+class ParticipationVoter extends Voter
+{
+    private RoleHierarchyInterface $roleHierarchy;
+
+    public function __construct(RoleHierarchyInterface $roleHierarchy)
+    {
+        $this->roleHierarchy = $roleHierarchy;
+    }
+
+    protected function supports(string $attribute, $subject): bool
+    {
+        return \in_array($attribute, ['PARTICIPATION_SHOW'])
+            && ($subject instanceof Participation);
+    }
+
+    /**
+     * @param Participation $subject
+     */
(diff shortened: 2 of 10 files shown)


src/Security/Voter/AdminRestrictionVoter.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\Security\Core\Authorization\Voter\Voter;
+use Symfony\Component\Security\Core\User\UserInterface;
+
+class AdminRestrictionVoter extends Voter
+{
+    protected function supports($attribute, $subject): bool
+    {
+        return \in_array($attribute, ['OWNER'])
+            && ($subject instanceof Talk || $subject instanceof Participation);
+    }
```
lyrixx, line 26: I don't really like the naming here
* AdminRestrictionVoter
   *  Not need for `Restriction` here. There is already a `Voter` Suffix
   * I don't get what `Admin` means here. "On what are you voting" should be the root of the naming. Someting Like ParticipationVoter or TalkVoter is better. So yes you'll have to split the voter in 2. Anyway, there is too many `if` in the `voteOnAttribute` method.
* `OWNER` is not a good attribute name. It's better to use an action like `TALK_EDIT`, `TALK_SHOW`, `PARTICIPATION_EDIT` etc. And yes, you'll have to split the voter in two voters :) But it's almost always better to have "fine grained permissions"

src/Security/Voter/AdminRestrictionVoter.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $user = $token->getUser();
+        if (!$user instanceof UserInterface) {
+            return false;
+        }
+
+        if ($subject instanceof Talk) {
+            if (\in_array($user, $subject->getUniqueUsers())) {
+                return true;
+            }
```
lyrixx, line 39: This code could be enchanced: in `$subject->getUniqueUsers()` you foreach all the record, the create a buffer. This consumer many memory for nothing. Let's move the foreach here a break as soon as the users is concerned. or you can move the code in the subject `return $suject->isUserAMember($user)`

config/packages/easy_admin/participation.yaml
```diff
@@ -7,6 +7,7 @@ easy_admin:
                 list: 'easy_admin/Participation/list.html.twig'
             list:
                 filters: ['conference', 'participant']
+                item_permission: ['OWNER', 'ROLE_ADMIN']
```
lyrixx, line 10: Let's encapsulate the decision about ROLE_ADMIN directly in the final voter.
This will avoid to scatter all permissions in the application

config/packages/easy_admin/participation.yaml
```diff
@@ -7,6 +7,7 @@ easy_admin:
                 list: 'easy_admin/Participation/list.html.twig'
             list:
                 filters: ['conference', 'participant']
+                item_permission: ['SHOW_PARTICIPATION']
```
lyrixx, line 10: You tried to be creative on this once :) 
But as I said in the previous comment, we prefer `PARTICIPATION_SHOW`. It namespace better attributes

src/Security/Voter/ParticipationVoter.php
```diff
@@ -0,0 +1,47 @@
...
+    protected function voteOnAttribute(string $attribute, $subject, TokenInterface $token): bool
+    {
+        $user = $token->getUser();
+        if (!$user instanceof UserInterface) {
+            return false;
+        }
+
+        if (\in_array('ROLE_ADMIN', $user->getRoles(), true)) {
+            return true;
+        }
```
lyrixx, line 39: You'll need to use the RoleHierarchy on this one. See https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/Security/Core/Authorization/Voter/RoleHierarchyVoter.php

lyrixx (commented):
this is better 👍🏼

welcoMattic (approved):
Looks great!

lyrixx (approved):
👍🏼 very nice 
Good job!

PR #121, opened 2021-02-10
Split conferences.tags column creation

welcoMattic:
The `NOT NULL` statement causes the fail of the migration. We have to wait for the whole data has been migrated before set the column not nullable.


lyrixx (approved):

lyrixx:
Thanks

PR #122, opened 2021-02-11
Fix Participation admin bug

HedicGuibert:
We currently have a bug in the admin if we try to access a conference that has participations. This is because the show action is currently not handled, so we're adding it.

I also allowed the show action for all conferences. It was not the case before but this might have been on purpose.

### DIFF
--- a/config/packages/easy_admin/conference.yaml
+++ b/config/packages/easy_admin/conference.yaml
@@ -15,9 +15,24 @@ easy_admin:
                     - { property: 'endAt', type: 'date' }
                     - { property: 'source', type: 'text' }
                     - { property: 'tags', type: 'array' }
+            show:
+                fields: &conferenceShow
+                    1: { property: name, type: text }
+                    2: { property: city, type: text }
+                    3: { property: country, type: country }
+                    4: { property: source, type: text }
+                    5: { property: description, type: text}
+                    6: { property: startAt, type: date }
+                    7: { property: endAt, type: date }
+                    8: { property: cfpUrl, type: url }
+                    9: { property: cfpEndAt, type: date }
+                    10: { property: siteUrl, type: url }
+                    11: { property: articleUrl, type: url }
+                    12: { property: submits, label: 'Submissions', type: embedded_list, css_class: "d-block mx-auto" }
+                    13: { property: participations, type: embedded_list, css_class: "d-block mx-auto" }
         NextConference:
             class: App\Entity\Conference
-            disabled_actions: ['show', 'delete']
+            disabled_actions: ['delete']
             list:
                 filters: ['name', 'city', 'country', 'startAt', 'tags', 'source']
                 actions:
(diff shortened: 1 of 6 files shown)


config/packages/easy_admin/conference.yaml
```diff
@@ -48,7 +63,22 @@ easy_admin:
                     - { property: 'cfpEndAt', type: 'date', 'type_options': { 'widget': 'single_text' } }
                     - { type: 'group', columns: 4, icon: 'cubes', label: 'Metadata' }
                     - { property: 'articleUrl', type: 'url' }
-        ExcludedConferences:
+            show:
+                fields:
+                    - { property: name, type: text }
```
lyrixx, line 68: too many duplication here. Let's use YAML merge operator: https://jolicode.com/blog/loperateur-de-merge-en-yaml-le-pire-ennemi-de-la-duplication-de-configuration

And thanks to it the list will be in the same order on all menu. Consistency is **really** important (in the UI, and in the code too: naming, style, etc)

config/packages/easy_admin/conference.yaml
```diff
@@ -48,7 +63,22 @@ easy_admin:
...
+                    - { property: city, type: text }
+                    - { property: source, type: text }
+                    - { property: description, type: text}
+                    - { property: startAt, type: date }
+                    - { property: endAt, type: date }
+                    - { property: cfpUrl, type: url }
+                    - { property: cfpEndAt, type: date }
+                    - { property: siteUrl, type: url }
+                    - { property: articleUrl, type: url }
+                    - { property: submits, label: 'Submissions', type: embedded_list, css_class: d-block mx-auto }
```
welcoMattic, line 79: Could you add `"` around the `css_class` values ? It will avoid unexpected YAML wrong parsing

lyrixx (commented):
Except one little comment 👍🏼 
You can merge yourself once fixed and green

templates/easy_admin/Participation/conference.html.twig
```diff
@@ -0,0 +1,8 @@
+<a href="{{ path('easyadmin', {
+        'entity': item.conference.guessConferenceType,
```
welcoMattic, line 2: I think you can retrieve the current entity class in the `easyadmin` object, and so the `guessConferenceType` could be removed

lyrixx: I don't think so. How easy admin could guess the **virtual** entity here?

welcoMattic: Indeed, I was pretty that it can retrieve it, but in fact, there are virtual entities. My bad

HedicGuibert: Unfortunately if I try to access the entity with `link_parameters.entity` it will always be "PassedConference", even if it should be a "NextConference" or "ExcludedConference".

But maybe there is an other way I didn't think about ?

lyrixx (approved):
Nice

Issue #1, opened 2018-02-23
Display a map with all location where we went at conferences


welcoMattic:
Diffrent types of markers:
- Conferences where we went as spectator
- Conferences where we went as speaker
- Conferences where we were sponsor (could be a star over other marker)

PR #125, opened 2021-02-15
Add map feature

HedicGuibert:
This references https://github.com/jolicode/starfleet/issues/1.

This adds a map to the homepage with a marker for every conference the team attended. Zoom is adjusted to conferences :

zoomed
Unzoomed

### DIFF
--- a/src/Command/UpdateConferencesCoordinatesCommand.php
+++ b/src/Command/UpdateConferencesCoordinatesCommand.php
@@ -1,0 +1,60 @@
+<?php
+
+/*
+ * This file is part of the Starfleet Project.
+ *
+ * (c) Starfleet <msantostefano@jolicode.com>
+ *
+ * For the full copyright and license information,
+ * please view the LICENSE file that was distributed with this source code.
+ */
+
+namespace App\Command;
+
+use App\Fetcher\LocationGuesser;
+use App\Repository\ConferenceRepository;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class UpdateConferencesCoordinatesCommand extends Command
+{
+    private ConferenceRepository $conferenceRepository;
+    private LocationGuesser $locationGuesser;
+    private EntityManagerInterface $em;
+
+    public function __construct(ConferenceRepository $conferenceRepository, LocationGuesser $locationGuesser, EntityManagerInterface $em)
+    {
+        $this->conferenceRepository = $conferenceRepository;
+        $this->locationGuesser = $locationGuesser;
+        $this->em = $em;
+
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this->setName('starfleet:conferences:update-coordinates');
+        $this->setDescription('Add coordinates to conferences already stored in the database.');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $conferences = $this->conferenceRepository->getNullCoordinatesConferences();
+        $updatedConferences = 0;
+
+        foreach ($conferences as $conference) {
+            $coordinates = $this->locationGuesser->getCoordinates($conference->getCity());
+            $conference->setCoordinates($coordinates);
+            ++$updatedConferences;
+
+            $this->em->flush();
+            $this->em->clear();
+        }
+
+        $output->writeln(sprintf('Updated conferences : %d', $updatedConferences));
+
+        return Command::SUCCESS;
+    }
+}
--- a/src/DataFixtures/FixtureBuilder.php
+++ b/src/DataFixtures/FixtureBuilder.php
@@ -117,6 +117,7 @@ class FixtureBuilder
             'cfpEndAt' => clone $date->modify('- 7 days'),
             'participations' => [],
             'online' => false,
+            'coordinates' => [self::getFaker()->longitude, self::getFaker()->latitude],
             'tags' => [
                 'PHP',
                 'DevOps',
@@ -139,6 +140,7 @@ class FixtureBuilder
         } else {
             $conference->setCity($description['city']);
             $conference->setCountry($description['country']);
+            $conference->setCoordinates($description['coordinates']);
         }
 
         foreach ($description['tags'] as $tag) {
--- a/src/Entity/Conference.php
+++ b/src/Entity/Conference.php
@@ -131,6 +131,13 @@ class Conference
      */
     private array $tags = [];
 
+    /**
+     * @ORM\Column(type="jsonb", nullable=true)
+     *
+     * @var array<float>|null
+     */
+    private ?array $coordinates = null;
+
     public function __construct()
     {
         $this->submits = new ArrayCollection();
@@ -403,5 +410,19 @@ class Conference
         }
 
         return 'NextConference';
+    }
+
+    /** @return array<float>|null */
+    public function getCoordinates(): ?array
+    {
+        return $this->coordinates;
+    }
+
+    /** @param array<float>|null $coordinates */
+    public function setCoordinates(?array $coordinates): self
+    {
+        $this->coordinates = $coordinates;
+
+        return $this;
     }
 }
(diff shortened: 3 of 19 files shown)


templates/conferences/list.html.twig
```diff
@@ -121,6 +121,7 @@
                     {{ futureConferences|length > 0 ? macro.display_conferences(futureConferences, 'future', activeTab) }}
                     {{ pastConferences|length > 0 ? macro.display_conferences(pastConferences, 'past', activeTab) }}
                 </div>
+            {% include "map.html.twig" %}
```
welcoMattic, line 124: Twig `include` tag is not recommended anymore, [Twig recommends to use the function instead](https://twig.symfony.com/doc/3.x/tags/include.html)

templates/map.html.twig
```diff
@@ -0,0 +1,7 @@
+{% if map_token is defined and map_token != 'your_mapbox_access_token' %}
```
welcoMattic, line 1: I would prefer to not test the value of the token. If someone forget to change the default token value, it's up to Mapbox to report the error, not to us. Moreover, there's no error message to tell the developer that they miss to change the token value.

src/Command/UpdateConferencesCoordinatesCommand.php
```diff
@@ -0,0 +1,61 @@
...
+{
+    private ConferenceRepository $conferenceRepository;
+    private LocationGuesser $locationGuesser;
+    private EntityManagerInterface $manager;
+
+    public function __construct(ConferenceRepository $conferenceRepository, LocationGuesser $locationGuesser, EntityManagerInterface $manager)
+    {
+        $this->conferenceRepository = $conferenceRepository;
+        $this->locationGuesser = $locationGuesser;
+        $this->manager = $manager;
```
lyrixx, line 31: We usually call it `em` or `entityManager`, but not `manger`

src/Command/UpdateConferencesCoordinatesCommand.php
```diff
@@ -0,0 +1,61 @@
...
+
+    protected function configure(): void
+    {
+        $this->setName('starfleet:conferences:update-coordinates');
+        $this->setDescription('Add coordinates to conferences already stored in the database.');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $allConferences = $this->conferenceRepository->findAll();
```
lyrixx, line 44: I would be better to use an iterator to no consume too much memory `$qb->getQuery()->iterate()`
Then, you'll need to flush + clear the EM in the loop

src/Command/UpdateConferencesCoordinatesCommand.php
```diff
@@ -0,0 +1,61 @@
...
+        $updatedConferences = 0;
+
+        foreach ($allConferences as $conference) {
+            if (!$conference->isOnline() && null === $conference->getCoordinates()) {
```
lyrixx, line 48: Us `if (XXX) { continue;}` to avoid too much indent

templates/map.html.twig
HedicGuibert: Sure. I chose to test the value because I thought it would be better to have it as an optionnal feature rather than a mandatory element. I'm changing it anyway.

src/DataFixtures/FixtureBuilder.php
```diff
@@ -117,6 +117,7 @@ public static function createConference(array $description = []): Conference
             'cfpEndAt' => clone $date->modify('- 7 days'),
             'participations' => [],
             'online' => false,
+            'coordinates' => [random_int(-50, 50), random_int(-50, 50)],
```
lyrixx, line 120: We have faker, may be it has something nice. Did you checked?

src/Entity/Conference.php
```diff
@@ -131,6 +131,13 @@ class Conference
      */
     private array $tags = [];
 
+    /**
+     * @ORM\Column(type="jsonb", nullable=true)
+     *
+     * @var array<float>
+     */
+    private ?array $coordinates = [];
```
lyrixx, line 139: I'm not a big fan of the typing you are using here.
* `?` says the value is nullable, but you never use this feature in the code. I guess it's for legacy data. So let's migrate legacy data from null to `[]`.
* but you initialize data with `[]`.

So now, there is 2 meanings to say data is empty [] and null.
You need to create something consistant.

I would go for : `NULL` in the DB and in the code, and an array not empty when there is something (so inizialize the data here with `null`)

src/Fetcher/ConfTechFetcher.php
```diff
@@ -132,6 +132,7 @@ private function denormalizeConference(array $rawConference, string $tag): ?Conf
             $query = sprintf('%s %s', $rawConference['city'], 'U.S.A.' === $rawConference['country'] ? 'United States of America' : $rawConference['country']);
             $continent = $this->locationGuesser->getContinent($query);
             $country = $this->locationGuesser->getCountry($query);
+            $coords = $this->locationGuesser->getCoordinates($query);
```
lyrixx, line 135: You do an extra call/request when the contient is null, or not enabled

lyrixx: BTW, I would move this block later, to avoid weird `/* @phpstan-ignore-next-line */`

templates/conferences/list.html.twig
lyrixx, line 124: wrong ident

templates/map.html.twig
lyrixx, line 1: This is ugly 😨 
Let's use a `null` default value instead

templates/map.html.twig
```diff
@@ -0,0 +1,7 @@
...
+    {% set apiConfig = {'type': 'FeatureCollection', 'features': [{'type': 'Feature', 'properties': {'marker-color': 'f7d325'}, 'geometry': {'type': 'MultiPoint', 'coordinates': conferenceCoordinates} }] } %}
+    {% set encode_apiConfig = apiConfig|json_encode|url_encode %}
```
lyrixx, line 3: Mix of snake case + camel case? Choose your side :) (camel case)

templates/map.html.twig
```diff
@@ -0,0 +1,7 @@
...
+    {% set url = 'https://api.mapbox.com/styles/v1/mapbox/streets-v11/static/geojson(%s)/auto/1000x600?access_token=%s'|format(encode_apiConfig, map_token) %}
```
lyrixx, line 4: This block a bit ugly. I would create a small and simple twig function for that.

src/Twig/MapboxUrlEncoder.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    public function getFunctions()
+    {
+        return [
+            new TwigFunction('mapboxEncode', [$this, 'encode']),
+        ];
+    }
+
+    /** @param array<array> $coordinates */
+    public function encode(array $coordinates, string $mapToken): string
```
lyrixx, line 27: Inject the token directly in this service and remove it from twig global

src/Twig/MapboxUrlEncoder.php
```diff
@@ -0,0 +1,44 @@
...
+                    'type' => 'Feature',
+                    'properties' => ['marker-color' => 'f7d325'],
+                    'geometry' => ['type' => 'MultiPoint', 'coordinates' => $coordinates],
+                ],
+            ],
+        ];
+        $encodeApiConfig = urlencode(json_encode($apiConfig));
+        $url = sprintf('https://api.mapbox.com/styles/v1/mapbox/streets-v11/static/geojson(%s)/auto/1000x600?access_token=%s', $encodeApiConfig, $mapToken);
+
+        return $url;
```
lyrixx, line 42: You can return the previous line directly

src/Repository/ConferenceRepository.php
```diff
@@ -85,6 +85,17 @@ public function getAllConferencesAsRow(): array
             ;
     }
 
+    /** @return array<Conference> */
+    public function getNullCoordinatesConferences(): array
+    {
+        return $this->createQueryBuilder('c')
+            ->andWhere('c.online = false')
```
welcoMattic, line 92: We should use a parameter here, since all RDBMS does not use the same values for boolean (i.e: pgsql accepts `False`, `false`, `0` at least)

src/Repository/ConferenceRepository.php
```diff
@@ -85,6 +85,18 @@ public function getAllConferencesAsRow(): array
...
 
+    /** @return array<Conference> */
+    public function getNullCoordinatesConferences(): array
+    {
+        return $this->createQueryBuilder('c')
+            ->andWhere('c.online = :falseValue')
+            ->setParameter('falseValue', false)
+            ->andWhere('c.coordinates IS NULL')
+            ->getQuery()
+            ->execute()
```
lyrixx, line 96: Can you update this query to not fetch all conference at once, it will consume too much memory.
Instead you must use an iterator `$qb->getQuery()->iterate()`

lyrixx (commented):
Only one little detail, and we'll be good 👍🏼

lyrixx (approved):
👍🏼 Thanks

Issue #123, opened 2021-02-11
Add Symfony Live conferences source


welcoMattic:
Based on this JSON, we can import all Symfony Live|World|Con conferences : https://live.symfony.com/api/conference/all.json

We need to add a new Fetcher for this source

PR #126, opened 2021-02-17
Symfony fetcher

HedicGuibert:
References https://github.com/jolicode/starfleet/issues/123 and adds a Fetcher for Symfony Conferences


src/Fetcher/SymfonyFetcher.php
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /** @param array<mixed> $rawConference */
+    private function denormalizeConference(array $rawConference): ?Conference
+    {
+        if (!$rawConference['is_online']) {
+            $city = $rawConference['city'];
+            $continent = $this->locationGuesser->getContinent($city);
+
+            if (!$continent->getEnabled() || !$continent instanceof Continent) {
```
lyrixx, line 65: This condition is a bit weird. If the left part is false, then the first part (which is the first to be executed) will generate a fatal error

src/Fetcher/SymfonyFetcher.php
```diff
@@ -0,0 +1,145 @@
...
+            }
+        }
+
+        $startDate = null;
+
+        if ($rawConference['starts_at']) {
+            $startDate = new \DateTimeImmutable($rawConference['starts_at']['date']);
+        }
+
+        // In case of invalid startDate, we skip the conference. It will be handled again later.
```
lyrixx, line 76: If the date is invalid, it will generate an exception, isn't?
So it looks like this is dead code (so you can remove it)

src/Fetcher/SymfonyFetcher.php
HedicGuibert: True.
This was some brainless c/c from what had been done before on other fetchers. I should update them all (same with what is said in next comment)

lyrixx (approved):
👍🏼

PR #128, opened 2021-02-17
Improve fetchers consistency

HedicGuibert:
We still had some weird parts in the fetchers that could be improved, so this is an attempt at fixing (some of) them.


lyrixx (approved):
👍🏼 good job

PR #129, opened 2021-02-17
New fetcher service

HedicGuibert:
This moves away from the command all the code that was doing stuff related to the DB or to the way we want to use the fetchers.

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -41,10 +41,9 @@ services:
         tags:
             - { name: twig.extension }
 
-    App\Command\FetchConferencesCommand:
+    App\Conferences\ConferencesHarvester:
         arguments:
-            - !tagged app.fetcher
-            - '@doctrine'
+            $fetchers: !tagged_iterator { tag: 'app.fetcher' }
 
     App\Security\Authenticator\GitHubAuthenticator:
         arguments:
--- a/src/Command/FetchConferencesCommand.php
+++ b/src/Command/FetchConferencesCommand.php
@@ -11,37 +11,19 @@
 
 namespace App\Command;
 
-use App\Entity\Conference;
-use App\Entity\ConferenceFilter;
-use App\Fetcher\FetcherInterface;
-use App\Repository\ConferenceFilterRepository;
-use App\Repository\ConferenceRepository;
-use App\Repository\FetcherConfigurationRepository;
-use Doctrine\Persistence\ManagerRegistry;
-use Doctrine\Persistence\ObjectManager;
+use App\Conferences\ConferencesHarvester;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\ConsoleOutputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Console\Style\SymfonyStyle;
 
 class FetchConferencesCommand extends Command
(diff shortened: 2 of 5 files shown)


src/Fetcher/FetcherDispatcher.php
```diff
@@ -0,0 +1,216 @@
...
+
+    public function __construct(FetcherConfigurationRepository $fetcherConfigurationRepository, ConferenceFilterRepository $conferenceFilterRepository, ConferenceRepository $conferenceRepository, EntityManagerInterface $em)
+    {
+        $this->fetcherConfigurationRepository = $fetcherConfigurationRepository;
+        $this->conferenceFilterRepository = $conferenceFilterRepository;
+        $this->conferenceRepository = $conferenceRepository;
+        $this->em = $em;
+    }
+
+    public function dispatch(FetcherInterface $fetcher): void
```
lyrixx, line 44: Can you find a better name for this class and method ?

By the way, I expected a class than handle all the fetchers at once, not one by one. I guess you might be confused about the output. You can remove the output from the command. It's a cron anyway.

src/Fetcher/FetcherDispatcher.php
```diff
@@ -0,0 +1,216 @@
...
+    private ConferenceFilterRepository $conferenceFilterRepository;
+    private ConferenceRepository $conferenceRepository;
+    private EntityManagerInterface $em;
+
+    /** @var array<ConferenceFilter> */
+    private array $conferenceFilters;
+    /** @var array<Conference> */
+    private array $existingConferences;
+
+    private int $updatedConferencesCount = 0;
```
lyrixx, line 33: No state in a service

config/services.yaml
```diff
@@ -41,10 +41,9 @@ services:
         tags:
             - { name: twig.extension }
 
-    App\Command\FetchConferencesCommand:
+    App\Fetcher\ConferencesHarvester:
         arguments:
-            - !tagged app.fetcher
-            - '@doctrine'
+            $fetchers: !tagged app.fetcher
```
lyrixx, line 46: `!$tagged` is deprecated IIRC, you must use `tagged_iterator` or `tagged_locator` (here you want the iterator)

src/Command/FetchConferencesCommand.php
```diff
@@ -58,171 +41,8 @@ protected function configure(): void
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $io = new SymfonyStyle($input, $output);
+        $this->conferencesHarvester->harvest($io);
```
lyrixx, line 44: This does not make sens to pass $io to a service. What will happen if you use this service from a web SAPI?

src/Fetcher/ConferencesHarvester.php
```diff
@@ -0,0 +1,222 @@
...
+/*
+ * This file is part of the Starfleet Project.
+ *
+ * (c) Starfleet <msantostefano@jolicode.com>
+ *
+ * For the full copyright and license information,
+ * please view the LICENSE file that was distributed with this source code.
+ */
+
+namespace App\Fetcher;
```
lyrixx, line 12: Not the right namespace IMHO.
The harvester **use** a fetcher, but it **isn't** a Fetcher.
I would use `Conference` NS instead

lyrixx (approved):
👍🏼  Top

PR #131, opened 2021-02-19
Handle null country

HedicGuibert:
If a conference country was set to null for some reason (like if someone manually added it) the flag would only show a broken img link, so in this case we just want to show nothing


templates/conferences/list.html.twig
```diff
@@ -18,6 +18,7 @@
                             <h6 class="card-subtitle mb-2 text-muted font-weight-normal">
                             {% if conference.online %}
                                 Online <img class="country-flag" alt="Online Icon" src="{{ asset('build/images/online.svg') }}">
+                            {% elseif conference.country == null %}
                             {% else %}
```
lyrixx, line 22:
```suggestion
                            {% elseif conference.country %}
```

same after

lyrixx (approved):

welcoMattic (approved):

PR #132, opened 2021-02-19
Fix conference update bug

HedicGuibert:
Conferences were never updating at all, this fixes it

### DIFF
--- a/src/Conferences/ConferencesHarvester.php
+++ b/src/Conferences/ConferencesHarvester.php
@@ -33,8 +33,6 @@ class ConferencesHarvester
     private iterable $fetchers;
     /** @var array<ConferenceFilter> */
     private array $conferenceFilters;
-    /** @var array<Conference> */
-    private array $existingConferences;
 
     /** @param iterable<FetcherInterface> $fetchers */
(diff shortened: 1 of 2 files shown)


src/Conferences/ConferencesHarvester.php
```diff
@@ -117,7 +118,8 @@ private function retrieveConferenceIfExists(Conference $conference): ?Conference
                     $existingConference['startAt']->format('Y-m-d') === $conference->getStartAt()->format('Y-m-d')
                     && $existingConference['endAt']->format('Y-m-d') === $conference->getEndAt()->format('Y-m-d')
                 ) {
-                    $matchedConference = $conference;
+                    $existingConference = $this->conferenceRepository->findOneBy(['id' => $existingConference['id']]);
+                    $matchedConference = $existingConference;
```
lyrixx, line 122: There is not break here, so we could fetch the "real" conference as late as possible to avoid extra SQL query

HedicGuibert: Fixed in https://github.com/jolicode/starfleet/pull/133

lyrixx:
Can you rebase please ?

lyrixx (approved):

Issue #97, opened 2020-12-16
The way we search for an existing conference does not scale


lyrixx:
ATM, we loop over all conference to find it a conference exist.
This does not scale.

https://github.com/jolicode/starfleet/blob/e91cc7d4c6e8feba193f056a2028bca1e1e2d93c/src/Command/FetchConferencesCommand.php#L101-L113

---

Please, find a way to avoid that. **But** before coding anything, please expose the algo (here if you're hot, or in slack it's fine)

lyrixx:
This part of the code is not a real bottle neck (tested with 10K conferences already in the DB)
https://blackfire.io/profiles/65638b7e-1b55-4dce-8a4b-3c417e66ce89/graph

PR #133, opened 2021-02-19
Improve levenshtein system

HedicGuibert:
This improves the way we are checking if a conference already exists.

Now we do this thanks to PGSql's levenshtein module rather than a php one.

This also helps with https://github.com/jolicode/starfleet/issues/97

### DIFF
--- a/src/Doctrine/ORM/Query/AST/Functions/Levenshtein.php
+++ b/src/Doctrine/ORM/Query/AST/Functions/Levenshtein.php
@@ -1,0 +1,44 @@
+<?php
+
+/*
+ * This file is part of the Starfleet Project.
+ *
+ * (c) Starfleet <msantostefano@jolicode.com>
+ *
+ * For the full copyright and license information,
+ * please view the LICENSE file that was distributed with this source code.
+ */
+
+namespace App\Doctrine\ORM\Query\AST\Functions;
+
+use Doctrine\ORM\Query\AST\Functions\FunctionNode;
+use Doctrine\ORM\Query\Lexer;
+use Doctrine\ORM\Query\Parser;
+use Doctrine\ORM\Query\SqlWalker;
+
+// Took from https://github.com/fza/mysql-doctrine-levenshtein-function#define-mysql-functions, thanks to him !
+class Levenshtein extends FunctionNode
+{
+    public ?object $firstStringExpression = null;
+    public ?object $secondStringExpression = null;
+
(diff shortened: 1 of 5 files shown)


src/Doctrine/ORM/Query/AST/Functions/Levenshtein.php
```diff
@@ -0,0 +1,43 @@
...
+    public ?object $firstStringExpression = null;
+    public ?object $secondStringExpression = null;
+
+    public function getSql(SqlWalker $sqlWalker)
+    {
+        return 'levenshtein('.
+            $this->firstStringExpression->dispatch($sqlWalker).', '.
+            $this->secondStringExpression->dispatch($sqlWalker).
+            ')';
+    }
```
lyrixx, line 31: The CS is a bit ugly, could you format it nicely? I guess on one line with sprintf will be OK

src/Repository/ConferenceRepository.php
```diff
@@ -101,6 +101,21 @@ public function getNullCoordinatesConferences(): ?\Generator
         }
     }
 
+    public function findExistingConference(Conference $conference): ?Conference
+    {
+        return $this->createQueryBuilder('c')
+            ->andWhere('c.startAt = :findStartAt AND c.endAt = :findEndAt')
```
lyrixx, line 107: it's a bit weird to name parameter `findXXX`, you can use `startAt`, `endAt`, and `name` instead.
BTW, you are using 2 different form with where AND. Could you use only one form?  (all condition in the same `andWhere` or 3 `andWhere`, up to you)

src/Repository/ConferenceRepository.php
```diff
@@ -101,6 +101,21 @@ public function getNullCoordinatesConferences(): ?\Generator
...
+            ->andWhere('levenshtein(c.name, :findName) < 4')
+            ->setParameters([
+                'findStartAt' => $conference->getStartAt(),
+                'findEndAt' => $conference->getEndAt(),
+                'findName' => $conference->getName(),
+            ])
+            ->getQuery()
```
lyrixx, line 114: This is almost impossible (I guess) but let's be defensive: `setMaxResults(1)`, otherwise you'll get an error when the query returns more than 1 result because of `getOneOrNullResult`

lyrixx (commented):
This is so nice 👍🏼 well done!

lyrixx (approved):

PR #134, opened 2021-02-22
Fix case when it's not possible to extract the continent from a conference 

HedicGuibert:


lyrixx:
@HedicGuibert I have update the PR title :)

lyrixx:
Thanks

PR #135, opened 2021-02-23
Improve map feature when few coordinates are provided

HedicGuibert:
This helps with cases where we have few coordinates in the database. Currently, if we have none (or if all conferences were online) we display a broken link. If we have only one conference to show on the map, it will show the street.

Now, we have a warning if we have no conferences to display, and if we have only one conference to display it adjusts the zoom.

### DIFF
--- a/templates/map.html.twig
+++ b/templates/map.html.twig
@@ -1,6 +1,8 @@
 {% set url = mapboxEncode(conferenceCoordinates) %}
 
-{% if url == null %}
+{% if conferenceCoordinates|length == 0 %}
+    <div class="text-muted">You don't have any conference to show on the map. This may be because they are all online.</div>
+{% elseif url == null %}
     <div class="alert alert-warning">It seems you forgot to provide an access token to enable the map feature.</div>
 {% else %}
     <img src="{{ url }}" alt="Map of where we attended conferences" class="w-100 mw-100">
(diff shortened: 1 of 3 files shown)


templates/map.html.twig
```diff
@@ -1,6 +1,12 @@
-{% set url = mapboxEncode(conferenceCoordinates) %}
+{% if conferenceCoordinates|length == 1 %}
+    {% set url = mapboxEncode(conferenceCoordinates, 5) %}
+{% else %}
+    {% set url = mapboxEncode(conferenceCoordinates) %}
+{% endif %}
```
lyrixx, line 5: This section could be simplified: the zoom "guess" could be done in the twig function

templates/map.html.twig
```diff
@@ -1,6 +1,12 @@
...
 
-{% if url == null %}
+{% if conferenceCoordinates|length == 0 %}
+    <div class="alert alert-warning">You don't have any conference to show on the map. This may be because they are all online.</div>
```
lyrixx, line 8: I really don't like the "alert" thing. We use "warning" or "danger" when the user does something wrong. here the user did not do anything wrong.

If you really want to put something, use a muted text

lyrixx (approved):

PR #136, opened 2021-02-23
Uppercase country code in the DB

HedicGuibert:
If we don't uppercase the country code in database this leads to a bug with symfony countryType not being able to recognize the country, so we should uppercase the country code in the database

### DIFF
--- a/migrations/Version20210223150819.php
+++ b/migrations/Version20210223150819.php
@@ -1,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Starfleet Project.
+ *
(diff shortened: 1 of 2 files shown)


src/Entity/Conference.php
```diff
@@ -357,7 +357,7 @@ public function getCountry(): ?string
 
     public function setCountry(?string $country): self
     {
-        $this->country = $country;
+        $this->country = strtoupper($country);
```
lyrixx, line 360: This will throw an error if `$country===null`, which is possible according to the signature.

And I would not to it here, but in the LocationGuesser Instead (and You already did it 👍🏼) so you can simply revert this :)

lyrixx (approved):
Thanks

PR #137, opened 2021-02-25
Add tests for all fetchers

HedicGuibert:
Since this is a major feature of the project, we should add tests to every fetcher.

I also made minor adjustments to some fetchers


lyrixx (approved):
👍🏼 thanks

PR #145, opened 2021-03-08
Add tests for ConferenceHarvester

HedicGuibert:
This adds tests for the ConferenceHarvester.

I also noticed 2 bugs that I fixed :

- If a fetcher wasn't active it would completely skip all remaining fetchers due to a return instead of continue
- `fnmmatch` wasn't working as intended and we weren't handling case issues for tags

### DIFF
--- a/src/Conferences/ConferencesHarvester.php
+++ b/src/Conferences/ConferencesHarvester.php
@@ -45,13 +45,13 @@ class ConferencesHarvester
         $this->logger = $logger ?: new NullLogger();
     }
 
-    public function harvest(): void
+    /** @return array<string,int> */
+    public function harvest(): array
     {
         $fetchersAmount = \count($this->fetchers);
         $currentFetcher = 0;
         $updatedConferencesCount = 0;
         $newConferencesCount = 0;
-
         foreach ($this->fetchers as $fetcher) {
             $this->logger->info(sprintf('Processing %d/%d fetchers : %s', ++$currentFetcher, $fetchersAmount, \get_class($fetcher)));
 
@@ -59,11 +59,10 @@ class ConferencesHarvester
             $fetcherConfiguration = $this->fetcherConfigurationRepository->findOneOrCreate($name);
 
             if (!$fetcherConfiguration->isActive()) {
-                return;
+                continue;
             }
 
             $config = $fetcherConfiguration->getConfiguration();
-
             foreach ($fetcher->fetch($config) as $conference) {
                 if ($this->shouldBeIgnored($conference)) {
                     continue;
@@ -86,6 +85,11 @@ class ConferencesHarvester
 
         $this->logger->notice($newConferencesCount.' newly added conference(s)');
         $this->logger->notice($updatedConferencesCount.' updated conference(s)');
+
+        return [
+            'newConferencesCount' => $newConferencesCount,
+            'updatedConferencesCount' => $updatedConferencesCount,
+        ];
     }
 
     private function updateExistingConference(Conference $existingConference, Conference $conference): bool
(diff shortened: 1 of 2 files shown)


src/Conferences/ConferencesHarvester.php
```diff
@@ -159,8 +163,8 @@ private function updateExistingConference(Conference $existingConference, Confer
     private function shouldBeIgnored(Conference $conference): bool
     {
         foreach ($this->getFilters() as $conferenceFilter) {
-            $filterName = $conferenceFilter->getName();
-            if (fnmatch($filterName, $conference->getName(), FNM_CASEFOLD) || \in_array($filterName, $conference->getTags())) {
+            $filterName = strtolower($conferenceFilter->getName());
+            if (str_contains(strtolower($conference->getName()), $filterName) || \in_array($filterName, array_map('strtolower', $conference->getTags()))) {
```
welcoMattic, line 167: `str_contains` has been introduced in PHP 8, we do not run on this version (yet ;) )

lyrixx: @welcoMattic symfony ships polyfill, we can use it.

src/Conferences/ConferencesHarvester.php
```diff
@@ -86,6 +85,11 @@ public function harvest(): void
 
         $this->logger->notice($newConferencesCount.' newly added conference(s)');
         $this->logger->notice($updatedConferencesCount.' updated conference(s)');
+
+        return [
+            'newConferences' => $newConferencesCount,
+            'updatedConferences' => $updatedConferencesCount,
```
lyrixx, line 91:
```suggestion
            'newConferencesCount' => $newConferencesCount,
            'updatedConferencesCount' => $updatedConferencesCount,
```

Even for testing, it's important to use a good name :)

tests/Conference/ConferenceHarvesterTest.php
```diff
@@ -0,0 +1,136 @@
...
+use Doctrine\ORM\EntityManager;
+use Prophecy\Argument;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+class ConferenceHarvesterTest extends KernelTestCase
+{
+    /**
+     * @dataProvider provideFetcherResponses
+     */
+    public function testFetch(array $fetcherData)
```
lyrixx, line 31: $fetcherData is a big black box. I would be better to use something like:

```suggestion
    public function testFetch(int $expectedUpdated, int $expectedAdded, string $name, bool isActive, XXXXX)
```

* Easier to read
* Benefit of type hint

tests/Conference/ConferenceHarvesterTest.php
```diff
@@ -0,0 +1,136 @@
...
+
+        $conferenceFilterRepository = $this->prophesize(ConferenceFilterRepository::class);
+        $conferenceFilterRepository
+            ->findAll()
+            ->willReturn([$filter = new ConferenceFilter()]);
+        $filter->setName('java');
+
+        $conferenceRepository = $this->prophesize(ConferenceRepository::class);
+        $conferenceRepository
+            ->findExistingConference(Argument::type('App\Entity\Conference'))
```
lyrixx, line 117:
```suggestion
            ->findExistingConference(Argument::type(Conference::class))
```

it's more natural :)

tests/Conference/ConferenceHarvesterTest.php
```diff
@@ -0,0 +1,136 @@
...
+            ->willReturn($fetcherData['existingConference']);
+
+        $entityManager = $this->prophesize(EntityManager::class);
+        $entityManager
+            ->persist(Argument::type('App\Entity\Conference'));
```
lyrixx, line 122: same here :)

lyrixx (commented):
It's very good! I let few minor comments

lyrixx:
Bravo. Thanks

PR #148, opened 2021-03-09
Allow adding a talk to multiple conferences

HedicGuibert:
Currently, we may add only one conference when we create a talk.

We should be able to create a submit for multiple conferences when we create a talk.


lyrixx (approved):

lyrixx:
Thanks

PR #150, opened 2021-03-12
Allow user to add authors on new Talk

HedicGuibert:
Right now, if we want to create a new talk given by multiple users, we have to create it, and then edit it to add new users, which is not handy. Same if an admin or anyone else wants to create a talk that will not be given by him, he has to create it then edit it.

Now we may add as many authors as we want, with the default author being current user. I made the field required however, but maybe we want to allow creating a talk with no author at all.

This will look like this :
talk

### DIFF
--- a/src/Controller/Admin/TalkController.php
+++ b/src/Controller/Admin/TalkController.php
@@ -30,8 +30,8 @@ class TalkController extends EasyAdminController
 
     protected function persistTalkEntity(Talk $talk, Form $newForm): void
     {
-        $conferences = $newForm->get('conference')->getData();
-
+        $conferences = $newForm->get('conferences')->getData();
+        $authors = $newForm->get('authors')->getData();
         $newSubmits = [];
 
         foreach ($conferences as $conference) {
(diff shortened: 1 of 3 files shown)


src/Form/TalkUserType.php
```diff
@@ -0,0 +1,40 @@
...
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Component\Security\Core\User\UserInterface;
+
+class TalkUserType extends AbstractType
+{
+    private UserInterface $user;
+
+    public function __construct(TokenStorageInterface $tokenStorage)
+    {
+        $this->user = $tokenStorage->getToken()->getUser();
```
lyrixx, line 27: You should not store the user but only the token storage here: No state in a service

src/Controller/Admin/TalkController.php
```diff
@@ -43,7 +43,10 @@ protected function persistTalkEntity(Talk $talk, Form $newForm): void
                 $submit->setConference($conference);
                 $submit->setTalk($talk);
                 $submit->setStatus(Submit::STATUS_PENDING);
-                $submit->addUser($this->getUser());
```
lyrixx, line 46: I saw this block of code in another PR an I don't get it. The form component should do it for you. What happens if you delete all the foreach?

lyrixx (approved):

src/Controller/Admin/TalkController.php
lyrixx: Talked in slack. It seems legit. I'll look at it later.

welcoMattic (approved):

PR #151, opened 2021-03-12
Show submits in menu

HedicGuibert:
If we want to see submits we can't do it directly from the menu and we have to go to conferences -> show a conference -> click `open in new tab` in the submits field.

This feels really weird so I suggest we add it to the menu, especially since everything has been made ready for it.

### DIFF
--- a/config/packages/easy_admin.yaml
+++ b/config/packages/easy_admin.yaml
@@ -14,6 +14,7 @@ easy_admin:
             - { entity: 'ConferenceFilter', label: 'Ignored Subjects', icon: 'times' }
             - { entity: 'User', label: 'Users', icon: 'users' }
             - { entity: 'Continent', label: 'Continents', icon: 'globe' }
+            - { entity: 'Submit', label: 'Submitted Talks', icon: 'check-circle' }
             - { label: 'Fetchers', route: 'fetcher_list', icon: 'bell' }
         assets:
             css:
--- a/config/packages/easy_admin/submit.yaml
+++ b/config/packages/easy_admin/submit.yaml
@@ -3,8 +3,13 @@ easy_admin:
         Submit:
             class: App\Entity\Submit
             disabled_actions: ['show']
+            edit:
+                role: SUBMIT_EDIT
+            delete:
+                role: SUBMIT_EDIT
             list:
                 filters: ['talk', 'conference', 'submittedAt', 'status', 'users']
+                role: ROLE_USER
                 fields:
                     - { property: 'talk' }
                     - { property: 'conference' }
@@ -12,6 +17,7 @@ easy_admin:
                     - { property: 'status', template: 'easy_admin/Submit/status.html.twig' }
                     - { property: 'users', type: 'array', label: 'Speakers' }
             form:
+                item_permission: SUBMIT_EDIT
                 fields:
                     - { type: 'group', columns: 6, icon: 'pencil', label: 'Submit' }
                     -   property: 'talk'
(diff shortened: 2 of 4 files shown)


config/packages/easy_admin.yaml
```diff
@@ -14,6 +14,7 @@ easy_admin:
             - { entity: 'ConferenceFilter', label: 'Ignored Subjects', icon: 'times' }
             - { entity: 'User', label: 'Users', icon: 'users' }
             - { entity: 'Continent', label: 'Continents', icon: 'globe' }
+            - { entity: 'Submit', label: 'Submits', icon: 'check-circle' }
```
welcoMattic, line 17: I prefer to display "Submitted talks" as label, instead of "Submits"

config/packages/security.yaml
```diff
@@ -35,3 +35,6 @@ security:
             - ROLE_USER
             - TALK_SHOW
             - PARTICIPATION_SHOW
+            - SUBMIT_EDIT_ALL
+        ROLE_USER:
+            - SUBMIT_EDIT_OWN
```
lyrixx, line 40: This is not really correct: instead of putting all the logic in the voter, you scatter it among the configuration and the application.

You should do the same thing as TALK_SHOW, PARTICIPATION_SHOW: used only `SUBMIT_EDIT`
And in the voter take some decision.

Finally, `SUBMIT_EDIT_ALL` seems not used.

src/Security/Voter/SubmitVoter.php
```diff
@@ -0,0 +1,57 @@
...
+
+    public function __construct(RoleHierarchyInterface $roleHierarchy)
+    {
+        $this->roleHierarchy = $roleHierarchy;
+    }
+
+    protected function supports(string $attribute, $subject): bool
+    {
+        return \in_array($attribute, ['SUBMIT_EDIT_OWN'])
+            && ($subject instanceof Submit);
```
lyrixx, line 32: no need for parenthesis

src/Security/Voter/SubmitVoter.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * @param Submit $subject
+     */
+    protected function voteOnAttribute(string $attribute, $subject, TokenInterface $token): bool
```
lyrixx, line 38:
```suggestion
    protected function voteOnAttribute(string $attribute, $submit, TokenInterface $token): bool
```

config/packages/easy_admin/submit.yaml
```diff
@@ -3,15 +3,21 @@ easy_admin:
...
                 filters: ['talk', 'conference', 'submittedAt', 'status', 'users']
+                role: ROLE_USER
                 fields:
                     - { property: 'talk' }
                     - { property: 'conference' }
                     - { property: 'submittedAt', type: 'date' }
                     - { property: 'status', template: 'easy_admin/Submit/status.html.twig' }
                     - { property: 'users', type: 'array', label: 'Speakers' }
             form:
+                item_permission: SUBMIT_EDIT
```
lyrixx, line 20: I'm not sure I'm commenting on the right line but in the list, we should filter in the DB, not in PHP

---

I noticed that during you presentation this noon. This is kind of wrong:
We retrieve from the DB many items, but we filter them after.
1. The pagination will be broken
2. We charge the DB / and the PHP application for nothing
=> A SQL filter must be set up to filter the data as soon as possible

But let's do that in another PR

lyrixx (approved):
Thanks. I left a comment but it can be do in another PR.

PR #152, opened 2021-03-12
Prevent slack notifications when a conference has no CfP url

HedicGuibert:
Prevents notifications from being sent if a conference has no CfP url.


lyrixx (approved):

lyrixx:
Thanks

PR #154, opened 2021-03-16
Remove Doctrine Events

HedicGuibert:
This completely removes Doctrine Events.

I also prevented notifications from being sent if a submit status is changed to done, because it wasn't handled well and it's probably not really interresting.


src/EventListener/SlackNotifierEventListener.php
```diff
@@ -117,4 +125,19 @@ public function onCfpEndingSoon(CfpEndingSoonEvent $event): void
         $payload['attachments'][] = $event->buildAttachment();
         $this->slackNotifier->notify($payload);
     }
+
+    /**
+     * @param GenericEvent<EasyAdminEvents> $event
```
lyrixx, line 130: Do this works? What does it mean?

src/EventListener/SlackNotifierEventListener.php
```diff
@@ -117,4 +125,19 @@ public function onCfpEndingSoon(CfpEndingSoonEvent $event): void
...
+    {
+        $entity = $event->getSubject();
+        if ($entity instanceof Submit) {
+            $statusChanged = $entity->getStatusChanged();
+            if (null !== $statusChanged) {
+                $this->eventDispatcher->dispatch(new SubmitStatusChangedEvent($entity));
+                $entity->resetStatusChanged();
+            }
+        }
+    }
```
lyrixx, line 142: This code does not belong to this class. It has nothing related to slack :)

src/EventListener/DoctrineEventSubscriber.php
```diff
@@ -1,88 +0,0 @@
...
-    }
-
-    public function postFlush(): void
-    {
-        if (1 === \count($this->conferencesAdded)) {
-            $this->eventDispatcher->dispatch(new NewConferenceEvent($this->conferencesAdded[0], $this->router));
-        } elseif (\count($this->conferencesAdded) > 1) {
-            $this->eventDispatcher->dispatch(new NewConferencesEvent($this->conferencesAdded, $this->router));
-        }
-    }
```
lyrixx, line 87: I failed to see where this code is now

src/EventListener/SlackNotifierEventListener.php
HedicGuibert: It works, it is just for phpstan. Since GenericEvent is an iterable, phpstan wants us to tell the iterable value type.
We could replace it with a phpstan-ignore-next-line but I think it is worse.

src/EventListener/DoctrineEventSubscriber.php
HedicGuibert: It is in the ConferenceHarvester, lines 98-102

src/EventListener/SlackNotifierEventListener.php
lyrixx: Ok ok 👍🏼 (you don't iterate on it, but not pb :) )

src/Entity/Submit.php
```diff
@@ -171,4 +174,14 @@ public function reduceSpeakersNames(): string
             return '' === $r ? $u->getName() : $r.', '.$u->getName();
         }, '');
     }
+
+    public function getStatusChanged(): bool
```
lyrixx, line 178:
```suggestion
    public function isStatusChanged(): bool
```
no?

src/Event/NewConferenceEvent.php
```diff
@@ -87,7 +85,7 @@ public function buildAttachment(): array
         if ($this->conference->getParticipations()->count() > 0) {
             $starfleetLinkField = SlackNotifier::SHORT_FIELD;
             $starfleetLinkField['title'] = 'Starfleet link  🚀';
-            $starfleetLinkField['value'] = $this->router->generate('conferences_show', [
+            $starfleetLinkField['value'] = $router->generate('conferences_show', [
```
lyrixx, line 88: Ohhhh, I see why the router was in this class before.
Don't modify this code for now, but It is absolutely not the place for this method :) 
It's not the responsibility to the event to know how to transform it to a slack attachment. It's the responsibility of the slack notifier.

TL;DR: don't do it now, but theses method should be move to the slack notifier (you can do it later, or not, up to you :) )

lyrixx (approved):
Very good job. Thanks

src/Entity/Submit.php
lyrixx: Could be done later, or not :)

PR #155, opened 2021-03-16
Fix conference_show route only working for attended conferences

HedicGuibert:
Currently, almost all slack links are not working because the conference_show route only works with attended conferences.

This will allow all conferences to work with the show_conference route.

### DIFF
--- a/src/Controller/ConferenceController.php
+++ b/src/Controller/ConferenceController.php
@@ -48,14 +48,8 @@ class ConferenceController extends AbstractController
     /**
      * @Route("/conferences/{slug}", name="conferences_show")
      */
-    public function showAction(string $slug, ConferenceRepository $conferenceRepository): Response
+    public function showAction(Conference $conference, ConferenceRepository $conferenceRepository): Response
     {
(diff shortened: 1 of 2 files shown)


welcoMattic (approved):

lyrixx (approved):
Thanks

src/Controller/ConferenceController.php
```diff
@@ -48,14 +48,8 @@ public function listByTagAction(string $tag, ConferenceRepository $conferenceRep
     /**
      * @Route("/conferences/{slug}", name="conferences_show")
      */
-    public function showAction(string $slug, ConferenceRepository $conferenceRepository): Response
+    public function showAction(Conference $conference, ConferenceRepository $conferenceRepository): Response
     {
-        $conference = $conferenceRepository->findOneAttended($slug);
```
lyrixx, line 53: This method is not use anymore. It must be deleted

lyrixx (changes requested):

lyrixx (approved):

PR #157, opened 2021-03-18
Refactor Slack Notifier

HedicGuibert:
This groups all the code relative to Slack in the Slack notifier service, as it should be.

I also translated all french messages to english and removed the `newConferenceEvent`


lyrixx (approved):
Very very good 👍🏼 
I did not review the SlackNotifier, since I guess you juste copied / pasted the code from the Events

HedicGuibert:
For the most part yes, I just translated a few things here and there and changed all the strings concatenation to `sprintf`

lyrixx:
Thanks

PR #166, opened 2021-04-13
Group slack notifications

HedicGuibert:
Fixes https://github.com/jolicode/starfleet/issues/163 and fixes https://github.com/jolicode/starfleet/issues/162

Still a WIP : we need to implement Slack interactions

Previously, we would send a daily notification with new conferences of the day that would be split in multiple messages. We would also send reminders for ending CFPs pretty often, which was spamming Slack a bit.
Now, we would only send one daily notification for both new conferences and CFPs, with the possibility of muting a conference if we don't want it to send ending CFP notifications later.

Since it also was suggested on Slack, I removed a lot of emojis, but we could easily adjust the design.

This refactors the notfication system and gets rid of the deprecated "attachments" system to use the new Slack Blocks.

If we merge, we will however need to update our Slack application to allow it to use user interaction, as it is needed for buttons to work. Other users of Starfleet will also need to do the same thing if they want it to work, so we probably should add it to the documentation.

How the daily notification would now look :
Daily Notif

How the notification for a new Talk would look :
New Talk

How the notification for a Submit status change would look :
Status Changed

PR #167, opened 2021-04-14
Help with manually added online conferences

HedicGuibert:
This will set the city to Online for online conferences that were added manually.

Previously someone could add a new conference, set it to be online, and leave the City field empty, which would mess with notifications : 
bug

I also renamed the ConferenceController to FrontController, which is a more accurate name IMO

### DIFF
--- a/src/Controller/FrontController.php
+++ b/src/Controller/FrontController.php
@@ -17,7 +17,7 @@ use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
 
-class ConferenceController extends AbstractController
+class FrontController extends AbstractController
 {
     /**
      * @Route("/", name="conferences_list")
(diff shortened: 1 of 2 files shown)


lyrixx (approved):

src/Controller/Admin/ConferenceController.php
```diff
@@ -0,0 +1,28 @@
...
+class ConferenceController extends EasyAdminController
+{
+    protected function persistNextConferenceEntity(Conference $conference, Form $newForm): void
+    {
+        if ($conference->isOnline() && null === $conference->getCity()) {
+            $conference->setCity('Online');
+        }
+
+        parent::persistEntity($conference);
+    }
```
lyrixx, line 27: Actually, I would made the change here: https://github.com/jolicode/starfleet/blob/master/src/Entity/Conference.php#L379

    public function setOnline(bool $online): self
    {
        $this->online = $online;
        $this->city = 'Online';
  
        return $this;
    }

lyrixx:
And the more I think about it, the less I agree with this change. If the conference is only, you must check in the code before calling `$conference->getCity()`

HedicGuibert:
The check for online conference will be handled by https://github.com/jolicode/starfleet/pull/166, it is probably not worth to change it here, but I agree the notifier should check if the conference is online.

lyrixx:
so, could you revert almost everything?

lyrixx (approved):

lyrixx:
Thanks

PR #168, opened 2021-04-22
Fix notification if all conferences are excluded

HedicGuibert:
Currently if all new conferences are exlcuded or don't have a CFP Url, there is no Slack message at all about conferences.

This should fix it and say "no conferences today".


lyrixx (approved):
Good catch 👍🏼

PR #174, opened 2021-05-03
Remove endDate check for same conferences and clean some code

HedicGuibert:
This fixes an issue when two sources add the same conference but don't set their end date to the same date. When this happens, we then have duplicates of this conference in the database.

Conferences links in Slack will now point to the front rather than to the admin


lyrixx (approved):
Thanks.

PR #175, opened 2021-05-04
Prevent empty string CFP URL from sending notifications

HedicGuibert:
Currently, some fetchers allow conferences to have an empty string as their CFP URL, and the notifier doesn't prevent empty CFP URL from sending notifications, which it should do.

### DIFF
--- a/tests/Notifiers/Slack/SlackNotifierTest.php
+++ b/tests/Notifiers/Slack/SlackNotifierTest.php
@@ -98,15 +98,6 @@ class SlackNotifierTest extends WebTestCase
             // One for $testConference, one to say no ending CFP.
             'expectedSections' => 2,
         ];
-
-        $noCfpUrlConference = $this->createConference(new \DateTime(), cfpUrl: null);
-
-        yield 'Test conference with no CFP URL is not sending notifications' => [
-            'newConferences' => [$testConference, $noCfpUrlConference],
-            'endingCfps' => [],
-            // One for $testConference, one to say no ending CFP.
-            'expectedSections' => 2,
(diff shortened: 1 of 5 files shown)


lyrixx:
I disagree with this move: If the CFP URL is missing, we should only remove the link to the CFP, not the entire notification

HedicGuibert:
I updated accordingly

tests/Notifiers/Slack/SlackNotifierTest.php
```diff
@@ -137,12 +128,12 @@ private function createSlackNotifier(array $endingCfps, HttpClientInterface $cli
...
     }
 
-    private function createConference(\DateTimeInterface $cfpEndAt, bool $excluded = false, ?string $cfpUrl = 'Test URL')
+    private function createConference(\DateTimeInterface $cfpEndAt, bool $excluded = false)
     {
         $conference = new Conference();
         $conference->setName('Test Conference');
         $conference->setCfpEndAt($cfpEndAt);
-        $conference->setCfpUrl($cfpUrl);
+        $conference->setCfpUrl('Test URL');
```
welcoMattic, line 136: It should be a valid URL here, no?

lyrixx: I would be better indeed!

HedicGuibert: I set a valid URL here

lyrixx (approved):
Thanks

welcoMattic (approved):
:+1:

PR #176, opened 2021-05-06
Fix empty notifications

HedicGuibert:
This should fix notifications from being sent if there is nothing to display

### DIFF
--- a/src/Controller/SlackController.php
+++ b/src/Controller/SlackController.php
@@ -55,8 +55,9 @@ class SlackController extends AbstractController
         $this->em->flush();
 
         $dailyConferences = $this->conferenceRepository->getDailyConferences();
+        $endingCfps = $this->conferenceRepository->getEndingCfpsByRemainingDays();
 
-        $blocks = $this->slackNotifier->buildDailyBlocks($dailyConferences);
+        $blocks = $this->slackNotifier->buildDailyBlocks($dailyConferences, $endingCfps);
 
         $body = [
             'replace_original' => true,
--- a/src/Notifiers/Slack/SlackNotifier.php
+++ b/src/Notifiers/Slack/SlackNotifier.php
@@ -14,39 +14,52 @@ namespace App\Notifiers\Slack;
 use App\Entity\Conference;
 use App\Entity\Submit;
 use App\Entity\Talk;
-use App\Repository\ConferenceRepository;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\RouterInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 class SlackNotifier
 {
+    private LoggerInterface $logger;
+
     public function __construct(
         private string $webHookUrl,
         private HttpClientInterface $httpClient,
         private RouterInterface $router,
-        private ConferenceRepository $conferenceRepository,
-        private SlackBlocksBuilder $slackBlocksBuilder
+        private SlackBlocksBuilder $slackBlocksBuilder,
+        ?LoggerInterface $logger = null,
     ) {
-    }
-
-    /** @param array<Conference> $newConferences */
-    public function sendDailyNotification(array $newConferences): void
-    {
-        $this->notify(['blocks' => $this->buildDailyBlocks($newConferences)]);
+        $this->logger = $logger ?: new NullLogger();
     }
 
     /**
-     * @param array<Conference> $newConferences
+     * @param array<Conference>            $newConferences
+     * @param array<int,array<Conference>> $endingCfps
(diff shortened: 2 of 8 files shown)


src/Notifiers/Slack/SlackNotifier.php
```diff
@@ -15,25 +15,41 @@
...
 use App\Repository\ConferenceRepository;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\RouterInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 class SlackNotifier
 {
+    private bool $hadNewConferences = true;
```
welcoMattic, line 26:
```suggestion
    private bool $hasNewConferences = true;
```

src/Notifiers/Slack/SlackNotifier.php
```diff
@@ -15,25 +15,41 @@
...
+    private bool $hadEndingCfps = true;
```
welcoMattic, line 27:
```suggestion
    private bool $hasEndingCfps = true;
```

src/Notifiers/Slack/SlackNotifier.php
welcoMattic: and for all usages

src/Notifiers/Slack/SlackNotifier.php
welcoMattic: and for all usages

src/Notifiers/Slack/SlackNotifier.php
```diff
@@ -15,25 +15,41 @@
...
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\RouterInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 class SlackNotifier
 {
+    private bool $hasNewConferences = true;
+    private bool $hasEndingCfps = true;
```
lyrixx, line 27: This class is a service. I must not have a state.
You should remove theses variables.

More over, you are computing too much things `buildConferencesBlocks` and `buildCfpsBlocks` in case there is nothing to display. The code could be simplified

src/Notifiers/Slack/SlackNotifier.php
```diff
@@ -168,24 +225,19 @@ private function buildConferencesBlocks(array $conferences): array
...
-        if (
-            0 === \count($endingCfps[0]) &&
-            0 === \count($endingCfps[1]) &&
-            0 === \count($endingCfps[5]) &&
-            0 === \count($endingCfps[10]) &&
-            0 === \count($endingCfps[20]) &&
-            0 === \count($endingCfps[30])
-        ) {
+        if (0 === \count($endingCfps)) {
             $cfpsBlock[] = $this->slackBlocksBuilder->buildSimpleSection('No ending CFP to display today !');
```
lyrixx, line 241: If there is nothing to display, we display nothing (not a test to say we have nothing to display).
Same in buildCfpsBlocks

src/Notifiers/Slack/SlackNotifier.php
```diff
@@ -168,29 +225,22 @@ private function buildConferencesBlocks(array $conferences): array
         ];
     }
 
-    /** @return array<array> */
-    private function buildCfpsBlocks(): array
+    /**
+     * @param array<mixed> $endingCfps
```
lyrixx, line 229: Mixed? conference instead?

src/Notifiers/Slack/SlackNotifier.php
```diff
@@ -15,37 +15,52 @@
...
+
+            return;
+        }
+
+        $this->notify(['blocks' => $this->buildDailyBlocks($newConferences, $endingCfps)]);
     }
 
     /**
      * @param array<Conference> $newConferences
+     * @param array<mixed>      $endingCfps
```
lyrixx, line 55: Mixed or conference?

src/Controller/SlackController.php
```diff
@@ -55,8 +55,9 @@ public function slackEndPoint(Request $request): Response
         $this->em->flush();
 
         $dailyConferences = $this->conferenceRepository->getDailyConferences();
+        $endingCfps = $this->slackNotifier->getEndingCfps();
 
-        $blocks = $this->slackNotifier->buildDailyBlocks($dailyConferences);
+        $blocks = $this->slackNotifier->buildDailyBlocks($dailyConferences, $endingCfps);
```
lyrixx, line 60: If we ignore everything, only a divider will be displayed, isn't?

lyrixx (approved):
Yeah, it's much better 👍🏼 Thanks

PR #183, opened 2021-05-11
Participation page

HedicGuibert:
This implements the user's participation page. It lacks fixtures however, because currently fixtures don't allow to see anything on the left side of the page.
If you want to  see something, you must be logged as admin, add a new participation to 2 different Next Conferences, then edit one of them to have its start date in the past.

I modified the EasyAdmin controllers a bit too, because I felt like the AdminController was not really relevant anymore. It is still there under the "ActionController" name, but IMO, we should get rid of it and move its methods to a "ConferenceController" since it is what it is doing now.

We should also wrap the past conference part somehow, because it will probably grow quite big with time

### DIFF
--- a/src/Controller/Admin/ParticipationController.php
+++ b/src/Controller/Admin/ParticipationController.php
@@ -69,12 +69,11 @@ class ParticipationController extends EasyAdminController
 
     #[Route(path: '/participation/cancel', name: 'participation_cancel')]
     #[IsGranted('ROLE_ADMIN')]
-    public function canceledAction(): Response
+    public function cancelledAction(): Response
     {
         /** @var Participation $participation */
         $participation = $this->request->attributes->get('easyadmin')['item'];
         $workflow = $this->workflowRegistry->get($participation, 'participation_request');
-
         $workflow->apply($participation, ParticipationTransition::CANCELLED);
 
         $this->em->flush();
--- a/src/Controller/Admin/ParticipationController.php
+++ b/src/Controller/Admin/ParticipationController.php
@@ -12,60 +12,34 @@
 namespace App\Controller\Admin;
 
 use AlterPHP\EasyAdminExtensionBundle\Controller\EasyAdminController;
-use App\Entity\Conference;
 use App\Entity\Participation;
 use App\Enum\Workflow\Transition\Participation as ParticipationTransition;
-use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Workflow\Registry as WorkflowRegistry;
 
-class AdminController extends EasyAdminController
+class ParticipationController extends EasyAdminController
 {
     public function __construct(
         private WorkflowRegistry $workflowRegistry,
     ) {
     }
 
-    public function excludeConferenceAction(): Response
+    protected function createNewParticipationEntity(): Participation
     {
-        /** @var Conference $conference */
-        $conference = $this->request->attributes->get('easyadmin')['item'];
-        $conference->setExcluded(true);
-
-        $this->em->flush();
-
-        return $this->redirectToRoute('easyadmin', [
-            'action' => 'list',
-            'entity' => $this->request->query->get('entity'),
-            'page' => $this->request->query->get('page'),
-        ]);
-    }
-
-    public function includeConferenceAction(): Response
-    {
-        /** @var Conference $conference */
-        $conference = $this->request->attributes->get('easyadmin')['item'];
-        $conference->setExcluded(false);
-
-        $this->em->flush();
-
-        return $this->redirectToRoute('easyadmin', [
-            'action' => 'list',
-            'entity' => $this->request->query->get('entity'),
-            'page' => $this->request->query->get('page'),
-        ]);
+        return new Participation($this->getUser());
     }
 
     #[Route(path: '/participation/accept', name: 'participation_accept')]
-    #[Security('ROLE_ADMIN')]
-    public function acceptAction(): Response
+    #[IsGranted('ROLE_ADMIN')]
+    public function acceptedAction(): Response
     {
         /** @var Participation $participation */
         $participation = $this->request->attributes->get('easyadmin')['item'];
         $workflow = $this->workflowRegistry->get($participation, 'participation_request');
 
-        $workflow->apply($participation, ParticipationTransition::ACCEPT);
+        $workflow->apply($participation, ParticipationTransition::ACCEPTED);
 
         $this->em->flush();
 
@@ -76,14 +50,14 @@ class AdminController extends EasyAdminController
     }
 
     #[Route(path: '/participation/reject', name: 'participation_reject')]
-    #[Security('ROLE_ADMIN')]
-    public function rejectAction(): Response
+    #[IsGranted('ROLE_ADMIN')]
+    public function rejectedAction(): Response
     {
         /** @var Participation $participation */
         $participation = $this->request->attributes->get('easyadmin')['item'];
         $workflow = $this->workflowRegistry->get($participation, 'participation_request');
 
-        $workflow->apply($participation, ParticipationTransition::REJECT);
+        $workflow->apply($participation, ParticipationTransition::REJECTED);
 
         $this->em->flush();
 
@@ -94,14 +68,14 @@ class AdminController extends EasyAdminController
     }
 
     #[Route(path: '/participation/cancel', name: 'participation_cancel')]
-    #[Security('ROLE_ADMIN')]
-    public function cancelAction(): Response
+    #[IsGranted('ROLE_ADMIN')]
+    public function canceledAction(): Response
     {
         /** @var Participation $participation */
         $participation = $this->request->attributes->get('easyadmin')['item'];
         $workflow = $this->workflowRegistry->get($participation, 'participation_request');
 
-        $workflow->apply($participation, ParticipationTransition::CANCEL);
+        $workflow->apply($participation, ParticipationTransition::CANCELLED);
 
         $this->em->flush();
 
--- a/src/Controller/User/UserController.php
+++ b/src/Controller/User/UserController.php
@@ -12,61 +12,37 @@
 namespace App\Controller\User;
 
 use App\Entity\Submit;
-use App\Entity\User;
 use App\Repository\ConferenceRepository;
+use App\Repository\ParticipationRepository;
 use App\Repository\SubmitRepository;
+use App\UX\UserChartBuilder;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
-use Symfony\UX\Chartjs\Builder\ChartBuilderInterface;
-use Symfony\UX\Chartjs\Model\Chart;
 
 class UserController extends AbstractController
 {
     public function __construct(
         private ConferenceRepository $conferenceRepository,
+        private ParticipationRepository $participationRepository,
         private SubmitRepository $submitRepository,
+        private UserChartBuilder $userChartBuilder,
     ) {
     }
 
     #[Route(path: '/user/account', name: 'user_account')]
-    public function userAccount(ChartBuilderInterface $chartBuilder): Response
+    public function userAccount(): Response
     {
-        /** @var User $user */
-        $user = $this->getUser();
-        $attendedConferences = $this->conferenceRepository->findAttentedConferencesByUser($user);
-        $pendingSubmits = range(1, 4); //$this->submitRepository->findUserSubmitsByStatus($user, Submit::STATUS_PENDING);
-        $acceptedSubmits = range(1, 8); //$this->submitRepository->findUserSubmitsByStatus($user, Submit::STATUS_ACCEPTED);
-        $rejectedSubmits = range(1, 14); //$this->submitRepository->findUserSubmitsByStatus($user, Submit::STATUS_REJECTED);
-        $upcomingSubmits = range(1, 1); //$this->submitRepository->findUserUpcomingUserSubmits($user);
+        $pastSubmits = $this->submitRepository->findUserSubmitsByStatus($this->getUser(), Submit::STATUS_DONE);
+        $futureSubmits = $this->submitRepository->findUserSubmitsByStatus($this->getUser(), Submit::STATUS_ACCEPTED);
+        $pastConferences = $this->conferenceRepository->findAttentedConferencesByUser($this->getUser());
+        $futureParticipations = $this->participationRepository->findFutureParticipationsByUser($this->getUser());
 
-        if (0 !== \count($pendingSubmits) + \count($acceptedSubmits) + \count($rejectedSubmits) + \count($upcomingSubmits)) {
-            $chart = $chartBuilder->createChart(Chart::TYPE_DOUGHNUT);
-            $chart->setData([
-                'labels' => ['Pending Submits', 'Accepted Submits', 'Rejected Submits', 'Upcoming Talks'],
-                'datasets' => [
-                    [
-                        'label' => 'Submitted talks',
-                        'data' => [\count($pendingSubmits), \count($acceptedSubmits), \count($rejectedSubmits), \count($upcomingSubmits)],
-                        'backgroundColor' => ['#ffc107', '#28a745', '#dc3545', '#007bff'],
-                    ],
-                ],
-            ]);
-
-            $chart->setOptions([
-                'legend' => [
-                    'position' => 'left',
-                    'align' => 'start',
-                ],
-            ]);
+        if (0 !== \count($pastSubmits) + \count($futureSubmits) + \count($pastConferences) + \count($futureParticipations)) {
+            $chart = $this->userChartBuilder->buildUserChart($pastSubmits, $futureSubmits, $pastConferences, $futureParticipations);
         }
 
         return $this->render('user/account.html.twig', [
-            'attendedConferences' => $attendedConferences,
-            'pendingSubmits' => $pendingSubmits,
-            'acceptedSubmits' => $acceptedSubmits,
-            'rejectedSubmits' => $rejectedSubmits,
-            'upcomingSubmits' => $upcomingSubmits,
             'chart' => $chart ?? null,
         ]);
     }
--- a/src/Controller/User/UserParticipationController.php
+++ b/src/Controller/User/UserParticipationController.php
@@ -18,13 +18,13 @@ use App\Repository\ConferenceRepository;
 use App\Repository\ParticipationRepository;
 use App\UX\UserChartBuilder;
 use Doctrine\ORM\EntityManagerInterface;
-use Exception;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
-use Symfony\Component\Security\Core\Exception\AccessDeniedException;
-use Symfony\Component\Workflow\WorkflowInterface;
+use Symfony\Component\Workflow\Exception\TransitionException;
+use Symfony\Component\Workflow\StateMachine;
 
 class UserParticipationController extends AbstractController
 {
@@ -33,7 +33,7 @@ class UserParticipationController extends AbstractController
         private ConferenceRepository $conferenceRepository,
         private UserChartBuilder $userChartBuilder,
         private EntityManagerInterface $em,
-        private WorkflowInterface $participationRequestStateMachine,
+        private StateMachine $workflow,
     ) {
     }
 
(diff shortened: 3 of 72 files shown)


src/Controller/Admin/ActionController.php
```diff
@@ -0,0 +1,49 @@
...
+ * please view the LICENSE file that was distributed with this source code.
+ */
+
+namespace App\Controller\Admin;
+
+use AlterPHP\EasyAdminExtensionBundle\Controller\EasyAdminController;
+use App\Entity\Conference;
+use Symfony\Component\HttpFoundation\Response;
+
+class ActionController extends EasyAdminController
```
welcoMattic, line 18:
```suggestion
class ConferenceActionController extends EasyAdminController
```

src/Controller/User/UserController.php
```diff
@@ -35,15 +39,20 @@ public function userAccount(ChartBuilderInterface $chartBuilder): Response
...
         $user = $this->getUser();
         $attendedConferences = $this->conferenceRepository->findAttentedConferencesByUser($user);
-        $pendingSubmits = range(1,4);//$this->submitRepository->findUserSubmitsByStatus($user, Submit::STATUS_PENDING);
-        $acceptedSubmits = range(1,8);//$this->submitRepository->findUserSubmitsByStatus($user, Submit::STATUS_ACCEPTED);
-        $rejectedSubmits = range(1,14);//$this->submitRepository->findUserSubmitsByStatus($user, Submit::STATUS_REJECTED);
-        $upcomingSubmits = range(1,1);//$this->submitRepository->findUserUpcomingUserSubmits($user);
+        $pendingSubmits = range(1, 4); //$this->submitRepository->findUserSubmitsByStatus($user, Submit::STATUS_PENDING);
+        $acceptedSubmits = range(1, 8); //$this->submitRepository->findUserSubmitsByStatus($user, Submit::STATUS_ACCEPTED);
+        $rejectedSubmits = range(1, 14); //$this->submitRepository->findUserSubmitsByStatus($user, Submit::STATUS_REJECTED);
+        $upcomingSubmits = range(1, 1); //$this->submitRepository->findUserUpcomingUserSubmits($user);
```
welcoMattic, line 45: it's fake data, you have to remove it

src/Controller/User/UserController.php
```diff
@@ -70,4 +79,27 @@ public function userAccount(ChartBuilderInterface $chartBuilder): Response
             'chart' => $chart ?? null,
         ]);
     }
+
+    #[Route(path: '/user/participation', name: 'user_participation')]
+    public function test(Request $request)
```
welcoMattic, line 84: test?

src/Controller/User/UserController.php
```diff
@@ -70,4 +79,27 @@ public function userAccount(ChartBuilderInterface $chartBuilder): Response
...
+    {
+        $participation = new Participation($this->getUser());
+        $form = $this->createForm(ParticipationType::class, $participation);
+
+        if ($form->handleRequest($request)->isSubmitted() && $form->isValid()) {
+            $this->em->persist($participation);
+            $this->em->flush();
+
+            return $this->redirectToRoute('user_participation');
```
welcoMattic, line 93: you redirect the user to the current route. No need to do that, just continue the execution of the action

src/Form/ParticipationType.php
```diff
@@ -0,0 +1,57 @@
...
+            ])
+            ->add('hotelStatus', ChoiceType::class, [
+                'choices' => Participation::HOTEL_STATUSES,
+                'required' => true,
+            ])
+            ->add('conferenceTicketStatus', ChoiceType::class, [
+                'choices' => Participation::CONFERENCE_TICKET_STATUSES,
+                'required' => true,
+            ])
+            ->add('submit', SubmitType::class)
```
welcoMattic, line 54: I don't like `SubmitType`, it's a UI element, it brings to the view, not to the FormType IMHO. Moreover, we don't need to check which button is clicked to submit the form.

templates/user/participation.html.twig
```diff
@@ -0,0 +1,32 @@
...
+                    {% for conference in pastConferences %}
+                        <div class="pl-5">
+                            <a href="{{ path('conferences_show', {slug: conference.slug}) }}">{{ conference.name }}</a>, starts at {{ conference.startAt|date('m/d/Y') }}
+                        </div>
+                    {% endfor %}
+                </div>
+            {% endif %}
+        </div>
+        <div class="col-6">
+            {% include "user/form/participation_form.html.twig" %}
```
welcoMattic, line 29: include tag is deprecated, you must use the include function instead

src/Controller/Admin/ActionController.php
lyrixx: Why **action**?

HedicGuibert: Action was probably not really an appropriate naming, but I don't think Conference is a good naming either because it's a bit confusing as we are using entity related in other places in the admin.

Maybe something like AdminController would be better ?

templates/user/participation.html.twig
HedicGuibert: It is for now, but I'm planning on using it again in the near future

src/Controller/Admin/ActionController.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        /** @var Conference $conference */
+        $conference = $this->request->attributes->get('easyadmin')['item'];
+        $conference->setExcluded(true);
+
+        $this->em->flush();
+
+        return $this->redirectToRoute('easyadmin', [
+            'action' => 'list',
+            'entity' => $this->request->query->get('entity'),
```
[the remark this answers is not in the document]
HedicGuibert: This is currently not possible because it may either be a NextConference or a PassedConference entity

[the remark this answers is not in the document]
lyrixx: okay 👍🏼

src/Controller/Admin/ActionController.php
HedicGuibert: I updated it, it's now still the ConferenceController but it is no longer the main EasyAdmin Controller

templates/user/participation.html.twig
HedicGuibert: Resolving and letting this untouched since it should be used soon

templates/user/account.html.twig
```diff
@@ -1,44 +1,41 @@
...
+            </a>
+        </li>
+        <li class="nav-item">
+            <a href="#" class="nav-link">
+                Edit my account
+            </a>
+        </li>
+        <li class="nav-item">
+            <a href="{{ app.user ? path('user_account') : path('login') }}" class="nav-link">
+                {{ app.user ? '➡️ Logout' }}
```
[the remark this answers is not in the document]
HedicGuibert: I'm not sure about this one, since I have a login text on my local instance, but I didn't look after it. Will do, letting this open for now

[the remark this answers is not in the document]
lyrixx: Where do it come from? 🤔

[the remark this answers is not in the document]
HedicGuibert: Ok so I didn't read the commentary well the first time. I thought we were talking about the header login button.

We don't have any login text on this button, but since the user must be logged in to access this page, we probably should leave it like this, no ?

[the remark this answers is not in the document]
lyrixx: there is an if (ternary) just above 🤔 I don't understand the code so

src/Controller/Admin/ParticipationController.php
```diff
@@ -76,7 +51,8 @@ public function acceptAction(): Response
     }
 
     #[Route(path: '/participation/reject', name: 'participation_reject')]
-    #[Security('ROLE_ADMIN')]
+
```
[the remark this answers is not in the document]
HedicGuibert: It seems fine but this is ugly, I will remove it

src/Controller/User/UserParticipationController.php
```diff
@@ -0,0 +1,98 @@
...
+        ]);
+    }
+
+    #[Route(path: '/user/participation_cancel/{id}', name: 'user_participations_cancel')]
+    public function cancelParticipation(Participation $participation): Response
+    {
+        $workflow = $this->workflowRegistry->get($participation, 'participation_request');
+
+        $workflow->apply($participation, ParticipationTransition::CANCEL);
+        $this->addFlash('info', 'Participation cancelled');
```
welcoMattic, line 76: it must be *after* the flush call

src/Entity/Participation.php
```diff
@@ -80,41 +71,30 @@ class Participation
     /**
      * @ORM\Column(type="string")
      */
-    #[Assert\Choice(choices: [
-        self::TRANSPORT_STATUS_NOT_NEEDED,
-        self::TRANSPORT_STATUS_NEEDED,
-        self::TRANSPORT_STATUS_BOOKED,
-    ])]
-    private string $transportStatus = self::TRANSPORT_STATUS_NOT_NEEDED;
+    #[Assert\Choice(choices: self::TRANSPORT_HOTEL_TICKETS_STATUSES)]
```
welcoMattic, line 74: please revert

src/Form/ParticipationType.php
```diff
@@ -0,0 +1,55 @@
...
+    ) {
+    }
+
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('conference', EntityType::class, [
+                'class' => Conference::class,
+                'choice_label' => 'name',
+                'query_builder' => $this->conferenceRepository->getFutureConferences(),
```
welcoMattic, line 36: Please suffix the method `getFutureConferences` with `QueryBuilder`, to avoid confusion about returning data or actual QueryBuilder

src/UX/UserChartBuilder.php
```diff
@@ -0,0 +1,129 @@
...
+        $chart->setData([
+            'labels' => [
+                sprintf('%d talks given', \count($pastSubmits)),
+                sprintf('%d talks you will give soon', \count($futureSubmits)),
+                sprintf('%d conferences you went to', \count($pastConferences)),
+                sprintf('%d conferences you will attend soon', \count($futureConferences)),
+            ],
+            'datasets' => [
+                [
+                    'label' => 'Submitted talks',
```
welcoMattic, line 43: Please use the Translator

welcoMattic: and everywhere it should be used ;) (no hard coded text)

templates/user/participation/more.html.twig
```diff
@@ -0,0 +1,84 @@
...
+            {% for participation in participations %}
+                <div class="row w-100">
+                    <a href="{{ path('conferences_show', {slug: participation.conference.slug}) }}">
+                        {{ participation.conference.name }}
+                    </a>, starts on {{ participation.conference.startAt|date('m/d/Y') }}
+                </div>
+            {% endfor %}
+        </div>
+    {% endif %}
+{% endblock %}
```
welcoMattic, line 84: Split this template in 4 templates, and make one action per "type" (which is more a filter than a type)

templates/user/submit/more.html.twig
```diff
@@ -0,0 +1,170 @@
...
+                        </a>
+                        <a href="{{ path('user_talks') }}" class="w-50">
+                            <div class="btn submit-pending p-0"><i class="fa fa-hourglass-half"></i> Pending</div>
+                        </a>
+                    </div>
+                </div>
+            {% endfor %}
+        </div>
+    {% endif %}
+{% endblock %}
```
welcoMattic, line 170: same as participation/more.html.twig

templates/user/user_layout.html.twig
```diff
@@ -0,0 +1,42 @@
...
+{% block body %}
+<div class="container-fluid bg-light">
+    <div class="container">
+        <div class="position-relative pt-5">
+            <div class="position-absolute lr-100">
+                {% for message in app.flashes('info') %}
+                    <div class="alert alert-info text-center">
+                        {{ message }}
+                    </div>
+                {% endfor %}
```
welcoMattic, line 17: It could be better to make a template we will include to handle all flashes types.

lyrixx: IMHO, this is fine for know. we'll see how to factorize code, only when needed

src/Controller/Admin/ParticipationController.php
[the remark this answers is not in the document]
HedicGuibert: It seems that the CS-fixer is adding these extra lines, so I probably should leave it like this, no ?

src/UX/UserChartBuilder.php
HedicGuibert: Will be done in a future PR when we upgrade to SF 5.3, so I'm resolving for now while leaving as is.

src/Form/ParticipationType.php
```diff
@@ -0,0 +1,62 @@
...
+
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('conference', DatalistType::class, [
+                'choices' => $this->conferenceRepository
+                    ->getFutureConferencesQueryBuilder()
+                    ->getQuery()
+                    ->execute(),
+                'mapped' => false,
```
lyrixx, line 38: We, this part seems wrong, cf the comment in the controller. You should attach a data transformer here
(Or create a new `ConferenceDataListType`, and put everything there)

src/Form/ParticipationType.php
```diff
@@ -0,0 +1,62 @@
...
+                    new NotBlank(),
+                    new NotNull(),
+                ],
+            ])
+            ->add('asSpeaker', CheckboxType::class, [
+                'required' => false,
+            ])
+            ->add('transportStatus', ChoiceType::class, [
+                'choices' => Participation::STATUSES,
+                'required' => true,
```
lyrixx, line 50: This is the default value, could be removed.

src/Form/ParticipationType.php
```diff
@@ -0,0 +1,62 @@
...
+                    ->execute(),
+                'mapped' => false,
+                'required' => true,
+                'constraints' => [
+                    new NotBlank(),
+                    new NotNull(),
+                ],
+            ])
+            ->add('asSpeaker', CheckboxType::class, [
+                'required' => false,
```
lyrixx, line 46: I think this is useless, isn't?

lyrixx (commented):
Good job 👍🏼 
I did not review the HTML / CSS part yet, since some modification on the PHP must be done before

lyrixx:
Oh, and I forgot: There is **many** new HTTP endpoint, but I saw 0 new tests. Please add at least one test for each new HTTP endpoint that ensure it's a 200 or 403. More assertions would be nice !

Note: If you are not sure if an HTTP endpoint will stay, you can avoid testing it, but if you are sure, please do.
Anyway, one will have to add many test before the merge in the main / master branch. Sonner the better 👼🏼

src/Controller/Admin/ParticipationController.php
```diff
@@ -12,53 +12,28 @@
...
-            'entity' => $this->request->query->get('entity'),
-            'page' => $this->request->query->get('page'),
-        ]);
+        return new Participation($this->getUser());
     }
 
     #[Route(path: '/participation/accept', name: 'participation_accept')]
-    #[Security('ROLE_ADMIN')]
+
+    /** @IsGranted("ROLE_ADMIN") */
```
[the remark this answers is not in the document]
HedicGuibert: The attribute is not working with security it seems.

The blank line is added by the CS-fixer, so I'm not sure about what to do with it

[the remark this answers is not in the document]
lyrixx: If it's the case, it's a bug in symfony and it must be reported
about the blank line: it's a bug in php cs fixer, and it must be reported too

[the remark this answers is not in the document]
HedicGuibert: After double checking :
The attribute is working fine (it was just not working with `#[Security]`)

The CS-fixer adds a blank line only if there is an annotation between the attribute and the method.
```
#[IsGranted('ROLE_ADMIN')]
/** @return Response */
```

This will add a blank line between them. This is however not the case if the return annotation is first.
Should I report it ?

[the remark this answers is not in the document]
lyrixx: > The attribute is working fine (it was just not working with `#[Security]`)

IMHO, this is because you misused the attritube. it must be something like `#[Security("is_granted('FOOBAR')")]`. Anyway



> This will add a blank line between them. This is however not the case if the return annotation is first.

yes you can, it will help other devs.
But in our case, just drop the `/** @return Response */`, and add a real type hint. PHP 7 FTW (I know, we are using 8.0)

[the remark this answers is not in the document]
HedicGuibert: It's already the case, it was just for testing purposes :)

src/Entity/Participation.php
```diff
@@ -81,40 +61,40 @@ class Participation
      * @ORM\Column(type="string")
      */
     #[Assert\Choice(choices: [
-        self::TRANSPORT_STATUS_NOT_NEEDED,
-        self::TRANSPORT_STATUS_NEEDED,
-        self::TRANSPORT_STATUS_BOOKED,
+        self::STATUS_NOT_NEEDED,
```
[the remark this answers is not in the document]
HedicGuibert: I did this but I had to revert because this was causing bugs, and since I couldn't use `array_values` either I had to keep things the way they were.

src/Controller/User/UserParticipationController.php
```diff
@@ -0,0 +1,153 @@
...
+            'chart' => $chart ?? null,
+        ]);
+    }
+
+    #[Route(path: '/user/participation-cancel/{id}', name: 'user_participations_cancel')]
+    public function cancelParticipation(Participation $participation): Response
+    {
+        if ($this->getUser() !== $participation->getParticipant()) {
+            throw new AccessDeniedException('You are not allowed to cancel someone else\'s participation.');
+        }
```
lyrixx, line 79: It should be handler by the security component, with a voter.

So I expect something like:

```
#IsGranted('PARTICIPATION_CANCEL', participation)
```
or in plain PHP, it's up to you

Because ATM, you are duplicating the logic everywhere you need it.
With a voter, the logic is only in the voter (and could be tested easilly)

src/Controller/User/UserParticipationController.php
```diff
@@ -0,0 +1,153 @@
...
+
+        try {
+            $this->workflow->apply($participation, ParticipationTransition::CANCELLED);
+        } catch (Exception $exception) {
```
lyrixx, line 83: Never ever use \Exception (BTW, we don't use `use Exception`).
It's better to catch only what you expect and let bubble other exceptions.
For exemple, if you forget a `;` somewhere in a listener, you will pass through the catch block, and it will be much harder to debug

src/DataTransformer/ConferenceNameTransformer.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+namespace App\DataTransformer;
+
+use App\Entity\Conference;
+use App\Repository\ConferenceRepository;
+use Symfony\Component\Form\DataTransformerInterface;
+use Symfony\Component\Form\Exception\TransformationFailedException;
+
+class ConferenceNameTransformer implements DataTransformerInterface
```
lyrixx, line 19: Perfect 👍🏼 this is much better

src/Form/ConferenceDatalistType.php
```diff
@@ -0,0 +1,47 @@
...
+        $resolver
+            ->setDefaults([
+                'choices' => $this->conferenceRepository
+                    ->getFutureConferencesQueryBuilder()
+                    ->getQuery()
+                    ->execute(),
+                'required' => true,
+                'constraints' => [
+                    new NotBlank(),
+                    new NotNull(),
```
lyrixx, line 38: NotBlank already covers NotNull. NotNull not needed IMHO

src/Form/ParticipationType.php
```diff
@@ -0,0 +1,51 @@
...
+            ])
+            ->add('conferenceTicketStatus', ChoiceType::class, [
+                'choices' => Participation::STATUSES,
+            ])
+        ;
+
+        $builder
+            ->get('conference')
+            ->addModelTransformer($this->transformer)
+        ;
```
lyrixx, line 49: Is it possible to attache the transformer directly on the ConferenceDatalistType ? (I'm not sure it's possible, but I think it is, since some core Symfony Type do that)

templates/user/account.html.twig
```diff
@@ -1,91 +1,40 @@
...
+            </li>
+            <li class="nav-item">
+                <a href="#" class="nav-link">
+                    Edit my account
+                </a>
+            </li>
+            <li class="nav-item">
+                <a href="{{ app.user ? path('user_account') : path('login') }}" class="nav-link">
+                    {{ app.user ? '➡️ Logout' }}
+                </a>
```
lyrixx, line 30: This is still weird ! if the user is NOT connected, the `<a>` contains nothing

lyrixx (commented):
Very nice. Some little things to fix and it will be OK

templates/user/account.html.twig
HedicGuibert: Sure, but this page is only accessible if you're logged in, so this probably shouldn't be an issue ? But I can add an if block, sure.

lyrixx: The you must simplify the href of the link

src/Security/Voter/UserAccountActionsVoter.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace App\Security\Voter;
+
+use App\Entity\Participation;
+use App\Entity\Submit;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Authorization\Voter\Voter;
+use Symfony\Component\Security\Core\User\UserInterface;
+
+class UserAccountActionsVoter extends Voter
```
lyrixx, line 20: Should be split in 2 voters `ParticipationVoter` and `SubmitVoter`

lyrixx (commented):
Very good 👍🏼

lyrixx (approved):

lyrixx:
Thanks a lot 👍🏼 💛

PR #194, opened 2021-06-11
User tests

HedicGuibert:
This improves the existing tests and add some more, as well as adding some more actions to the submits (edit and remove). I'm not 100% convinced about the remove one, any opinion on this ?

We still miss a few tests however (for the remove and edit submit actions)

### DIFF
--- a/assets/js/tomSelect.js
+++ b/assets/js/tomSelect.js
@@ -1,0 +1,5 @@
+const TomSelect = require('tom-select');
+
+[...document.getElementsByClassName('tom-select')].forEach(select =>
+    new TomSelect(select)
+);
--- a/src/Controller/User/UserSubmitController.php
+++ b/src/Controller/User/UserSubmitController.php
@@ -147,4 +147,36 @@ class UserSubmitController extends AbstractController
             'submits' => $this->submitRepository->findUserSubmitsByStatus($this->getUser(), Submit::STATUS_REJECTED),
         ]);
     }
+
+    #[IsGranted('SUBMIT_ACTION', 'submit')]
+    #[Route(path: '/user/submit-edit/{id}', name: 'edit_submit')]
+    public function editSubmit(Submit $submit, Request $request): Response
+    {
+        $form = $this->createForm(SubmitType::class, $submit);
+
+        if ($form->handleRequest($request)->isSubmitted() && $form->isValid()) {
+            $this->em->persist($submit);
+            $this->em->flush();
+
+            $this->addFlash('info', 'Your submit has been submitted.');
+
+            return $this->redirectToRoute('user_submits');
+        }
+
+        return $this->render('/user/submit/submit_edit.html.twig', [
+            'submit' => $submit,
+            'form' => $form->createView(),
+        ]);
+    }
+
+    #[IsGranted('SUBMIT_ACTION', 'submit')]
+    #[Route(path: '/user/submit-remove/{id}', name: 'remove_submit')]
+    public function removeSubmit(Submit $submit): Response
+    {
+        $this->em->remove($submit);
+        $this->em->flush();
+        $this->addFlash('info', 'Submit removed from the database.');
+
+        return $this->redirectToRoute('user_submits');
+    }
 }
--- a/src/Controller/User/UserSubmitController.php
+++ b/src/Controller/User/UserSubmitController.php
@@ -116,10 +116,10 @@ class UserSubmitController extends AbstractController
         return $this->redirectToRoute('user_submits');
     }
 
-    #[Route(path: '/user/future-submits', name: 'future_submits')]
+    #[Route(path: '/user/accepted-submits', name: 'accepted_submits')]
     public function futureSubmits(): Response
     {
-        return $this->render('user/submit/future_submits.html.twig', [
+        return $this->render('user/submit/accepted_submits.html.twig', [
             'submits' => $this->submitRepository->findUserSubmitsByStatus($this->getUser(), Submit::STATUS_ACCEPTED),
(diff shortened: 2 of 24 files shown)


templates/user/submit/submit_edit.html.twig
```diff
@@ -0,0 +1,67 @@
...
+
+{% form_theme form 'bootstrap_4_layout.html.twig'%}
+{% form_theme form.conference _self %}
+{% form_theme form.users _self %}
+
+{% block _submit_conference_widget %}
+    <input list="{{ id }}" id="{{ id }}_input" class="form-control" name="{{ full_name }}" autocomplete="off" placeholder="{{ value }}">
+    <datalist id="{{ id }}">
+        {% for choice in choices %}
+            <option value="{{ choice.name }}"></option>
```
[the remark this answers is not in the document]
HedicGuibert: Unfortunately wen can't, because the value of the options is visible

src/Controller/User/UserSubmitController.php
```diff
@@ -147,4 +147,36 @@ public function rejectedSubmits(): Response
             'submits' => $this->submitRepository->findUserSubmitsByStatus($this->getUser(), Submit::STATUS_REJECTED),
         ]);
     }
+
+    #[IsGranted('SUBMIT_ACTION', 'submit')]
```
lyrixx, line 151: Are you sure you should pass a string? Does it work?

templates/user/participation/future_participations.html.twig
```diff
@@ -8,12 +8,12 @@
 {% endblock %}
 
 {% block content %}
-    <div class="row mb-5">
+    <div class="row mb-5" id="future-participations-block">
```
lyrixx, line 11: the id is useless?

templates/user/submit/submit.html.twig
```diff
@@ -17,6 +17,14 @@
     <label for="submits_users" class="d-block">Users</label>
 {% endblock %}
 
+{% block _submit_users_widget %}
+    <select name="submit[users][]" id="tom-select" multiple>
```
lyrixx, line 21: there is a twig variable to get the `name`
And the id is needed? IE this custom ID for the connection between the label and the select

templates/user/submit/submit_edit.html.twig
```diff
@@ -0,0 +1,71 @@
...
+        {% endfor %}
+    </datalist>
+{% endblock %}
+
+{% block _submit_users_label %}
+    <label for="submits_users" class="d-block">Users</label>
+{% endblock %}
+
+{% block _submit_users_widget %}
+    <select name="submit[users][]" id="tom-select" multiple>
```
lyrixx, line 21: same here

templates/user/submit/submit_edit.html.twig
```diff
@@ -0,0 +1,71 @@
...
+    <input list="{{ id }}" id="{{ id }}_input" class="form-control" name="{{ full_name }}" autocomplete="off" placeholder="{{ value }}">
+    <datalist id="{{ id }}">
+        {% for choice in choices %}
+            <option value="{{ choice.name }}"></option>
+        {% endfor %}
+    </datalist>
+{% endblock %}
+
+{% block _submit_users_label %}
+    <label for="submits_users" class="d-block">Users</label>
```
lyrixx, line 17: the `for` seems wrong so

src/Controller/User/UserSubmitController.php
HedicGuibert: Yes it does, and this is the only way I found, I could invest further however

templates/user/participation/future_participations.html.twig
HedicGuibert: It's used in the tests

src/Controller/User/UserSubmitController.php
lyrixx: ok ok, 👍🏼 no need to dig

templates/user/submit/submit.html.twig
HedicGuibert: This ID is used to transform the select into a tom select, we need a unique ID for this

lyrixx: The ID is always unique :)

HedicGuibert: I know, but since it seems a bit complicated to use a class, select them all and transform them all into a tomselect, I decided to instead use an ID and limit the number of tomselects on a page to 1.

You think  we should use a class ?

lyrixx: I'm not against a class, but I don't see what's hard here :/ You can have the same behavior with a class. You can add the class to only one element. More over, it's more flexible

assets/js/tomSelect.js
```diff
@@ -0,0 +1,5 @@
+const TomSelect = require('tom-select');
+
+[...document.getElementsByClassName('tom-select')].forEach(select =>
```
lyrixx, line 3: why don't you use:

```suggestion
document.getElementsByClassName('tom-select').forEach(select =>
```

lyrixx (commented):
👍🏼 except one little comment

assets/js/tomSelect.js
HedicGuibert: It's not possible to forEach on a HTMLCollection, which is what `getElementsByClassName` returns. We could use a for/of instead however, but for some reason it throws errors in the console.
We also could use a QuerySelector if you don't like what I did here.

lyrixx: ok ok :)

PR #197, opened 2021-06-23
Fix notification bug when updating a submit through a talk

HedicGuibert:
If someone tries tu update a submit through its talk, the notification isn't fired, so this fixes it.


lyrixx (approved):
Seems good (but I'm not sure :p)

PR #198, opened 2021-06-24
Add cron to update past submits

HedicGuibert:
This will add a command that should be a cron to automatically set the status of accepted submits to done if the conference is over.
It will also add a reminder for the user if they have a pending submit for a conference which is over.

This is to prevent users from forgetting to change the marking of their conferences and leaving irrelevant entries in the database.

### DIFF
--- a/src/Command/ConferenceStatusesCommand.php
+++ b/src/Command/ConferenceStatusesCommand.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the Starfleet Project.
+ *
+ * (c) Starfleet <msantostefano@jolicode.com>
+ *
+ * For the full copyright and license information,
+ * please view the LICENSE file that was distributed with this source code.
+ */
+
+namespace App\Command;
+
+use App\Repository\SubmitRepository;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class ConferenceStatusesCommand extends Command
+{
+    public function __construct(
+        private SubmitRepository $submitRepository,
+    ) {
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this->setName('starfleet:submits:statuses');
+        $this->setDescription('Updates the statuses of past submits');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $this->submitRepository->updateDoneSubmits();
+
+        return Command::SUCCESS;
+    }
+}
--- a/src/Repository/SubmitRepository.php
+++ b/src/Repository/SubmitRepository.php
@@ -37,7 +37,7 @@ class SubmitRepository extends ServiceEntityRepository
             ->andWhere('s.status = :status')
             ->setParameter('status', $status)
             ->innerJoin('s.conference', 'c')
-            ->orderBy('c.createdAt', 'ASC')
+            ->orderBy('s.createdAt', 'DESC')
             ->getQuery()
             ->execute()
         ;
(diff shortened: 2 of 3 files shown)


src/Command/ConferenceStatusesCommand.php
```diff
@@ -0,0 +1,43 @@
...
+
+    protected function configure(): void
+    {
+        $this->setName('starfleet:submits:statuses');
+        $this->setDescription('Updates the statuses of past submits');
+    }
+
+    /**
+     * @param ConsoleOutputInterface $output
+     */
```
lyrixx, line 36: To remove

src/Repository/SubmitRepository.php
```diff
@@ -54,6 +54,60 @@ public function findUserSubmits(User $user): array
...
+    {
+        $em = $this->getEntityManager();
+
+        foreach ($this->getDoneSubmits() as $submit) {
+            $submit->setStatus(Submit::STATUS_DONE);
+            $submit->resetStatusChanged();
+        }
+
+        foreach ($this->getOldPendingSubmits() as $submit) {
+            $em->remove($submit);
```
lyrixx, line 67: Do we really want to **remove** things?

Anyway, if yes, **one** `delete` SQL query would be much more performant and a SELECT + many DELETE

src/Repository/SubmitRepository.php
```diff
@@ -54,6 +54,56 @@ public function findUserSubmits(User $user): array
         ;
     }
 
+    public function updateDoneSubmits(): void
+    {
+        $em = $this->getEntityManager();
+
+        foreach ($this->getDoneSubmits() as $submit) {
```
lyrixx, line 61: Instead of a select + many update, 
use a single update instead

lyrixx:
Don't forgot to comment on the PR when it's ready. I almost missed your update on this one :)

lyrixx:
ping @HedicGuibert

src/Repository/SubmitRepository.php
```diff
@@ -54,6 +54,65 @@ public function findUserSubmits(User $user): array
...
+                'status_accepted' => Submit::STATUS_ACCEPTED,
+                'today' => $today,
+            ])
+            ->getQuery()
+            ->execute()
+        ;
+    }
+
+    /** @return array<Submit> */
+    public function getOldPendingSubmits(): array
```
lyrixx, line 98: Dead code?

tests/Repository/SubmitRepositoryTest.php
```diff
@@ -0,0 +1,28 @@
...
+class SubmitRepositoryTest extends KernelTestCase
+{
+    public function testUpdateDoneSubmitsWork()
+    {
+        static::bootKernel();
+
+        $submitRepository = static::$container->get(SubmitRepository::class);
+        $submitRepository->updateDoneSubmits();
+
+        self::assertSame(0, \count($submitRepository->getDoneSubmits()));
```
lyrixx, line 26: You are hydrating all the DB, for just a count. Your DB can return the value directly. 

And the method `getDoneSubmits` is not used in the app, so you can inline the query here

You really must remember this rule: 

> You must not hydrate data when it's not really necessary ! It kills 🐻

 😄

tests/Repository/SubmitRepositoryTest.php
```diff
@@ -0,0 +1,45 @@
...
+
+namespace App\Tests\Repository;
+
+use App\Entity\Submit;
+use App\Repository\SubmitRepository;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+class SubmitRepositoryTest extends KernelTestCase
+{
+    public function testUpdateDoneSubmitsWork()
```
lyrixx, line 20: This test does not really work: You have no guarantee that some `pastAcceptedSubmitsCount` is > 0 before the test start. So this test does not prove values has been updated.

But I guess it works with fresh fixtures, so it kindy works in the CI but not in dev. 

Anyway, I'm gonna merge this PR. But this is something you have to think about

1. what are my initial data (you must be sure what they are, and you can't rely on fixtures only, since another tests may have modified it -- or you need a strong policy about fixuere 
2. do something
3. assert it's OK

Some ref:
* https://jolicode.com/blog/ameliorer-la-dx-de-vos-fixtures-php
* http://agileinaflash.blogspot.com/2009/02/first.html

lyrixx (approved):

tests/Repository/SubmitRepositoryTest.php
HedicGuibert: This test wasn't really made to test that values have been updated, but rather to ensure that after running the command, no submits are left with an accepted status if their conference has ended.

But you're right, since we may have 0 submits in that case due to fixtures, the tests could pass while the command is actually broken, so I'll improve this test and make sure it actually updates some data. And since I wrote several tests that rely on fixtures data, I will make an issue to improve this.

lyrixx: > This test wasn't really made to test that values have been updated, but rather to ensure that after running the command, no submits are left with an accepted status if their conference has ended.

I should have :)

PR #203, opened 2021-08-16
Talks page

HedicGuibert:
Still in progress

This adds the missing talks page, but this also changes the design we use in all the user account (mainly moving from a custom list ordering to some cards).
Some parts of the design still need improvments though.

I also made some improvments here and there, some will need some more work (thinking of new flash for instance).

A few pictures to see what it looks like :
talks
show
subits


src/Form/EventSubscribers/CreateTalkSubscriber.php
```diff
@@ -0,0 +1,45 @@
...
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Form\FormEvent;
+use Symfony\Component\Form\FormEvents;
+
+class CreateTalkSubscriber implements EventSubscriberInterface
+{
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            FormEvents::POST_SUBMIT => 'onPostSubmit',
```
lyrixx, line 24: Should occurs after the validation isn't ?

src/Form/NewTalkType.php
```diff
@@ -0,0 +1,40 @@
...
+            ->add('intro', TextareaType::class)
+            ->add('conference', ConferenceDatalistType::class, [
+                'mapped' => false,
+            ])
+            ->add('users', EntityType::class, [
+                'class' => User::class,
+                'multiple' => true,
+                'mapped' => false,
+            ])
+            ->addEventSubscriber(new CreateTalkSubscriber())
```
lyrixx, line 37: Should use dependency injection (but see the other comment before, since everything will be moved to this class)

src/Repository/TalkRepository.php
```diff
@@ -27,4 +29,24 @@ public function __construct(ManagerRegistry $registry)
...
+    public function findUserTalks(User $user): array
+    {
+        return $this->createUserQueryBuilder($user)
+            ->orderBy('t.id', 'DESC')
+            ->getQuery()
+            ->execute()
+        ;
+    }
+
+    private function createUserQueryBuilder(User $user): QueryBuilder
```
lyrixx, line 43: It's only used once. I think it does not deserve to create a new method

templates/user/submit/accepted_submits.html.twig
```diff
@@ -1,65 +1,58 @@
...
-                <div class="row mb-2 w-75">
-                    {{ submit.talk.intro }}
-                </div>
+                <div class="card-body">
+                    {% if submit.users|length > 1 %}
+                        <div>
+                            Given with :
+                            {% for user in submit.users %}
+                                {% if app.user != user %}
+                                    {{ user.name }}{% if not loop.last %}, {% endif %}
```
lyrixx, line 27: the loop.last may not work because of the if. You must filter the collection in the [for](https://twig.symfony.com/doc/3.x/filters/filter.html)

templates/user/talk/talk.html.twig
```diff
@@ -0,0 +1,94 @@
...
+                            <div class="justify-content-center mt-2">
+                                <a href="{{ path('edit_talk', { id: talk.id }) }}" class="action-edit ml-4">
+                                    <div class="btn color-edit p-0"><i class="fa fa-edit"></i> Edit</div>
+                                </a>
+                            </div>
+                        </div>
+                    </div>
+                    {% if loop.last and talks|length > 3 %}
+                        <a href="{{ path('user_talks_all') }}">...Show more</a>
+                    {% endif %}
```
lyrixx, line 59: Could be move outside the for

templates/user/talk/talk.html.twig
```diff
@@ -0,0 +1,94 @@
...
+        {{ form_start(form, {
+            attr: {
+                class: 'row flex-column',
+            }
+        }) }}
+            {{ form_row(form.title) }}
+            {{ form_row(form.intro) }}
+            {{ form_row(form.conference) }}
+            {{ form_row(form.users, {
+                value: [app.user]
```
lyrixx, line 77: should be done in PHP, not in the template

templates/user/talk/talk_edit.html.twig
```diff
@@ -0,0 +1,35 @@
...
+{% endblock %}
+
+{% block content %}
+    <div class="row flex-column align-items-center w-100">
+        <h2>Edit your talk</h2>
+        <div class="row flex-column align-items-center w-50">
+            {{ form_start(form, {
+                attr: {
+                    class: 'row flex-column w-100',
+                    id: form.vars.id
```
lyrixx, line 24: why this line?

HedicGuibert: If I don't add this, the form doesn't have an ID and I need it to have one beacause I use it in the tests

templates/user/talk/talk.html.twig
HedicGuibert: Not completely sure about the fix I did in the UserTalkController

src/Controller/User/UserTalkController.php
```diff
@@ -0,0 +1,117 @@
...
+        private EntityManagerInterface $em,
+    ) {
+    }
+
+    #[Route('/user/talks', name: 'user_talks')]
+    public function userTalks(Request $request): Response
+    {
+        $talk = new Talk();
+        $form = $this->createForm(NewTalkType::class, $talk);
+        $form->get('users')->setData([$this->getUser()]);
```
lyrixx, line 41: It would be better to set the data in the talk directly, than passing via the form. Something like:

```
$talk = new Talk();
$task->setUsers([$this->getUser()])
```

With that, you can remove this line, and the one in the template

HedicGuibert: Unfortunately, the user is not set on the talk directly, but on the submit, this is why I couldn't do that

lyrixx: Discussed IRL. the type is decoupled from the model. So this is fine 👍🏼

templates/user/talk/talk_edit.html.twig
lyrixx: That's weird 🤔  but ok for now

HedicGuibert:
I just squashed

PR #204, opened 2021-08-17
Improve map

HedicGuibert:
Faker generates wrong latitudes for mapbox, this fixes it.
It also changes the theme of the map.


lyrixx (approved):
Thanks

PR #205, opened 2021-08-18
Add User Profile page

HedicGuibert:
This adds the Edit Profile page to the user account, as well as improving small things (constraints, and links color)

### DIFF
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -50,6 +50,9 @@ class User implements UserInterface, \Serializable
 
     /**
      * @ORM\Column(name="email", type="string", length=255, unique=true)
+     *
+     * @Assert\NotBlank()
+     * @Assert\Email()
      */
     private string $email;
 
--- a/src/Form/NewTalkType.php
+++ b/src/Form/NewTalkType.php
@@ -20,6 +20,7 @@ use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Form\FormEvent;
 use Symfony\Component\Form\FormEvents;
+use Symfony\Component\Validator\Constraints\NotBlank;
 
 class NewTalkType extends AbstractType
 {
@@ -30,11 +31,17 @@ class NewTalkType extends AbstractType
             ->add('intro', TextareaType::class)
             ->add('conference', ConferenceDatalistType::class, [
                 'mapped' => false,
+                'constraints' => [
+                    new NotBlank(),
+                ],
             ])
             ->add('users', EntityType::class, [
                 'class' => User::class,
                 'multiple' => true,
                 'mapped' => false,
+                'constraints' => [
+                    new NotBlank(),
+                ],
             ])
             ->addEventListener(
                 FormEvents::POST_SUBMIT,
--- a/src/Form/UserAccount/UserProfileType.php
+++ b/src/Form/UserAccount/UserProfileType.php
@@ -1,0 +1,126 @@
+<?php
+
+/*
+ * This file is part of the Starfleet Project.
+ *
+ * (c) Starfleet <msantostefano@jolicode.com>
+ *
+ * For the full copyright and license information,
+ * please view the LICENSE file that was distributed with this source code.
+ */
+
+namespace App\Form\UserAccount;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\Extension\Core\Type\EmailType;
+use Symfony\Component\Form\Extension\Core\Type\PasswordType;
+use Symfony\Component\Form\Extension\Core\Type\TextareaType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\Form\FormBuilderInterface;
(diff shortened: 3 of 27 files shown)


src/Form/NewTalkType.php
```diff
@@ -30,11 +32,19 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             ->add('intro', TextareaType::class)
             ->add('conference', ConferenceDatalistType::class, [
                 'mapped' => false,
+                'constraints' => [
+                    new NotNull(),
```
lyrixx, line 36: NotBlank is enough

src/Form/UserAccount/UserProfileType.php
```diff
@@ -0,0 +1,89 @@
...
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('name', TextType::class, [
+                'required' => false,
+            ])
+            ->add('email', EmailType::class, [
+                'constraints' => [
+                    new NotNull(message: 'You must have an email address.'),
+                    new NotBlank(message: 'You must have an email address.'),
```
lyrixx, line 35: * The validation should be on the Entity (or DTO when it exists) not on the form.
* and let the default message
* and it miss an Email constraint

src/Form/UserAccount/UserProfileType.php
```diff
@@ -0,0 +1,89 @@
...
+                ],
+            ])
+            ->add('password', PasswordType::class, [
+                'required' => false,
+            ])
```
lyrixx, line 40: Hmm, not sur ! 

1. you don't ask for the previous password
2. you don't encode the new password in the controller
3. if the password is empty, it erase the previous password

=> Usually, we create another form / endpoint / template to deal with password

templates/user/user_layout.html.twig
```diff
@@ -47,9 +47,11 @@
                                 <div class="btn btn-secondary">Participations</div>
                             </a>
                         {% endif %}
-                        <a href="{{ path('user_account') }}">
-                            <div class="btn btn-secondary">Edit Account</div>
-                        </a>
+                        {% if 'profile' not in app.request.get('_route') %}
```
lyrixx, line 50: This is fragile

src/Entity/User.php
```diff
@@ -50,6 +50,9 @@ class User implements UserInterface, \Serializable
 
     /**
      * @ORM\Column(name="email", type="string", length=255, unique=true)
+     *
+     * @Assert\NotBlank
+     * @Assert\Email
```
lyrixx, line 55:
```suggestion
     * @Assert\NotBlank()
     * @Assert\Email()
```

🙏🏼

src/Form/UserAccount/UserProfileType.php
```diff
@@ -0,0 +1,121 @@
...
+                            return $form->get('previousPassword')->addError(new FormError('Invalid password.'));
+                        }
+
+                        if (!$this->encoder->isPasswordValid($user, $previousPasswordData)) {
+                            return $form->get('previousPassword')->addError(new FormError('Invalid password.'));
+                        }
+
+                        if (\strlen($passwordData) < 7) {
+                            return $form->get('password')->addError(new FormError('Your password must have at least 6 characters.'));
+                        }
```
lyrixx, line 66: I would use: https://symfony.com/doc/current/reference/constraints/UserPassword.html

src/Form/UserAccount/UserProfileType.php
```diff
@@ -0,0 +1,121 @@
...
+
+                        $user->setPassword($this->encoder->encodePassword($user, $passwordData));
+                    }
+                },
+                1
+            )
```
lyrixx, line 72: You'll have to run this listener after the validation.

And usually, we register listeners after form child type.

And I would use a dedicated method

templates/user/user_layout.html.twig
```diff
@@ -23,32 +23,26 @@
...
-                        {% endif %}
-                        {% if 'participation' not in app.request.get('_route') %}
-                            <a href="{{ path('user_participations') }}">
-                                <div class="btn btn-secondary">Participations</div>
-                            </a>
-                        {% endif %}
-                        <a href="{{ path('user_account') }}">
-                            <div class="btn btn-secondary">Edit Account</div>
+                        <a href="{{ path('user_submits') }}">
+                            <div class="btn btn-{{ path('user_submits') == app.request.requestUri ? 'primary' : 'secondary' }}">Submits</div>
```
lyrixx, line 31: fragile too :) 

There is two way to solve that

1. app.request.attribute.get('_route') ==
2. declare a variable `section` (for example) in this template, and override in all subtemplate.

Usually, we do 2, because it allow to highlite a link on many different pages

For example, on 
* https://jolicode.com/blog/
* https://jolicode.com/blog/jai-teste-tailwind-css

=> Two different actions, two different template, but the same menu is highlighted

src/Form/UserAccount/UserProfileType.php
HedicGuibert: Unfortunately this Constraint prevents us from leaving the field empty, and we don't want to force the user to update his password

lyrixx: okay :+1:

lyrixx (approved):
Thanks

PR #210, opened 2021-08-26
Improve user account

HedicGuibert:
This improves the user account in various ways :

- Adds the submittedBy field to the Submit entity. We previously didn't know who created a submit, and since a submit may have several users, it is useful to know who created it, especially since we want to know this in the future when we impement notifications.
- Improve design and fix few things in templates
- Adds new constraints


lyrixx (commented):
👍🏼 Thanks

PR #213, opened 2021-08-27
Improve tests using Foundry

### LINKED ISSUE #211: [User Account] Improve Tests
Currently, the user account tests rely too much on fixtures. If someone changes the fixtures they may break.
They could also have unexpected behaviour.

We should make sure they work on their own.
This should take 1 day, 2 at most.

HedicGuibert:
Fix #211 

Some tests were in the `functionnal` directory, which I didn't like, so I moved them and created some new directories.
As a consequence I also had to move 1 or 2 Controllers that were at the root of the Controller directory to match these changes.


src/Validator/Constraints/UniqueTitleValidator.php
```diff
@@ -0,0 +1,45 @@
...
+    ) {
+    }
+
+    /**
+     * @param string      $formTitle
+     * @param UniqueTitle $constraint
+     */
+    public function validate($formTitle, Constraint $constraint): void
+    {
+        if (!$formTitle || '' === $formTitle) {
```
lyrixx, line 31:
```suggestion
        if (!$formTitle) {
```

src/Controller/Security/SecurityController.php
```diff
@@ -9,7 +9,7 @@
  * please view the LICENSE file that was distributed with this source code.
  */
 
-namespace App\Controller;
+namespace App\Controller\Security;
```
lyrixx, line 12: It's the security for the front, isn't? So i should goes to the `Front` dir?

lyrixx (commented):
It's nice, but most of the PR is not related to "
Improve tests using Foundry" :(

src/Controller/Security/SecurityController.php
HedicGuibert: It's used to login users so I don't know, and since the authenticators are in App\Security it looks ok to have it in App\Controller\Security since they are related ?

I'm not sure however.

lyrixx: Actually, I don't really understand the new directory structure.
I would split in 3 dir:
* admin
* front
* slack
so if this controller is used to auth use on the frontend, it must go to the front dir

src/Validator/Constraints/UniqueTitleValidator.php
HedicGuibert: These changes were made in a previous PR and this one was not yet rebased, this is fixed in #210

lyrixx:
Can you rebase? Thanks

src/Controller/Security/SecurityController.php
HedicGuibert: I moved from Controller\Security to Controller\Front but I will leave the other ones as they are for now since this is not really what this PR should do.

I will probably open a new PR dedicated to the Controllers howewer, as their naming could also use some improvments :).

lyrixx (approved):
👍🏼

PR #214, opened 2021-08-30
Improve Controllers naming and directory structure

HedicGuibert:
Our Controllers directory structure and naming could be improved, `User` is not really accurate , we have Controllers at the root of the Controller directory, and we repeat "User" in the names of our Controllers.

I updated the Tests names to match the Controllers ones.


lyrixx (commented):
Love it, thanks

welcoMattic (approved):

PR #215, opened 2021-09-02
Remove dead code

HedicGuibert:
I noticed dead code il the User entity, we should remove it


HedicGuibert:
Not relevant as I didn't notice that we implemented the Serializable interface

PR #216, opened 2021-09-03
Notifications

### LINKED ISSUE #212: [User Account] Add notifications
Users currently don't know if one of their submit or participation was accepted. They have to go and check their submits and participations, and see if one had its status changed.

They also don't know if someone ele created submitted a talk with them included.

And they don't know when a featured conference is added (this feature is currently not implemented however but it will be soon).

We should add a notifications system so uthers get inputs about what is hapenning for them. Other notification ideas welcome :).
This should take 4 days.

HedicGuibert:
Fix #212 

This PR implements notifications for Starfleet users. 
Users should be notified when a new Featured Conference is added, when someone submits a talk for them, or when someone changes the status of one of their submit or participation. Some more may come in the future.


assets/js/highlight_scroll.js
```diff
@@ -0,0 +1,15 @@
+const url = new URL(window.location.href);
+const id = url.searchParams.get("highlight");
```
lyrixx, line 2: Let's move that to the inner function (once loaded)

assets/js/highlight_scroll.js
```diff
@@ -0,0 +1,15 @@
...
+
+window.addEventListener('load', function(e) {
+    if (id) {
```
lyrixx, line 5: early return please

src/Controller/UserAccount/NotificationController.php
```diff
@@ -0,0 +1,51 @@
...
+
+class NotificationController extends AbstractController
+{
+    public function __construct(
+        private EntityManagerInterface $em,
+        private NotificationRepository $notificationRepository,
+    ) {
+    }
+
+    #[IsGranted('READ_NOTIFICATION', 'notification')]
```
lyrixx, line 22:
```suggestion
    #[IsGranted('NOTIFICATION_READ', 'notification')]
```

we use most common thing first, then more specialized

src/Controller/UserAccount/NotificationController.php
```diff
@@ -0,0 +1,51 @@
...
+    #[Route(path: '/user/read-notification/{id}', name: 'read_notification', methods: ["POST"])]
+    public function readNotification(Notification $notification, Request $request): Response
+    {
+        if (!$this->isCsrfTokenValid('', $request->request->get('token'))) {
+            throw new AccessDeniedException();
+        }
+
+        $notification->setRead(true);
```
lyrixx, line 30:
```suggestion
        $notification->markAsRead();
```

and remove the setter on notification.

src/Controller/UserAccount/NotificationController.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->em->flush();
+        $this->addFlash('info', 'Notification marked as read.');
```
lyrixx, line 32: This message is a bit raw, isn't?

HedicGuibert:
This is a picture of what it looks like for now since we made a first design attempt. 
Advices are welcome :)

notifications

src/Controller/UserAccount/NotificationController.php
```diff
@@ -0,0 +1,51 @@
...
+
+        return $this->redirectToRoute($request->query->get('route_name'));
```
lyrixx, line 34: * What if the route name has parameters? 
* what if the routename is missing from the $request?

You must pass the current URL, or pass all attributes + request param

src/Controller/UserAccount/NotificationController.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    #[Route(path: '/user/read-all-notification', name: 'read_all_notification', methods: ["POST"])]
+    public function readAllNotification(Request $request): Response
+    {
+        if (!$this->isCsrfTokenValid('', $request->request->get('token'))) {
+            throw new AccessDeniedException();
+        }
+
+        $this->notificationRepository->markAllAsRead($this->getUser());
```
lyrixx, line 44:
```suggestion
        $this->notificationRepository->markAllAsReadForUser($this->getUser());
```

src/Entity/Conference.php
```diff
@@ -11,10 +11,10 @@
 
 namespace App\Entity;
 
-use Doctrine\Common\Collections\ArrayCollection;
-use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
 use Gedmo\Mapping\Annotation as Gedmo;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\Common\Collections\ArrayCollection;
 use Gedmo\Timestampable\Traits\TimestampableEntity;
```
lyrixx, line 18: Why all theses diff? It's not alpha sorted anymore.

Did you run php-cs-fixer ?

src/Entity/Notifications/NewFeaturedConferenceNotification.php
```diff
@@ -0,0 +1,30 @@
...
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity()
+ */
+class NewFeaturedConferenceNotification extends Notification
+{
+    /**
+     * @ORM\ManyToOne(targetEntity=Conference::class)
+     * @ORM\JoinColumn
```
lyrixx, line 16:
```suggestion
     * @ORM\JoinColumn()
```

src/Entity/Notifications/Notification.php
```diff
@@ -0,0 +1,119 @@
...
+
+use App\Entity\User;
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\ORM\Mapping\InheritanceType;
+use Doctrine\ORM\Mapping\DiscriminatorMap;
+use Doctrine\ORM\Mapping\DiscriminatorColumn;
+use Symfony\Component\Validator\Constraints as Assert;
+
+/**
+ * @ORM\Entity
```
lyrixx, line 13:
```suggestion
 * @ORM\Entity()
```

(on each PR I have to write this comment 🙈)

src/Entity/Notifications/Notification.php
```diff
@@ -0,0 +1,119 @@
...
+ * @InheritanceType("SINGLE_TABLE")
+ * @DiscriminatorColumn(name="type", type="string")
+ * @DiscriminatorMap({
+ *      "submitAdded" = "SubmitAddedNotification",
+ *      "submitStatusChanged" = "SubmitStatusChangedNotification",
+ *      "participationStatusChangedNotification" = "ParticipationStatusChangedNotification",
+ *      "newFeaturedConference" = "NewFeaturedConferenceNotification"
+ * })
+ */
+abstract class Notification
```
lyrixx, line 23:
```suggestion
abstract class AbstractNotification
```

src/Entity/Notifications/Notification.php
```diff
@@ -0,0 +1,119 @@
...
+     * @ORM\GeneratedValue
+     * @ORM\Column(type="integer")
+     */
+    protected int $id;
+
+    /**
+     * @ORM\ManyToOne(targetEntity=User::class, inversedBy="notifications")
+     * @ORM\JoinColumn(nullable=false)
+     */
+    protected User $targetUser;
```
lyrixx, line 49: why `targetUser` ? `user` is enought, isn't?

src/Entity/Notifications/Notification.php
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * @ORM\Id
+     * @ORM\GeneratedValue
+     * @ORM\Column(type="integer")
+     */
+    protected int $id;
+
+    /**
+     * @ORM\ManyToOne(targetEntity=User::class, inversedBy="notifications")
+     * @ORM\JoinColumn(nullable=false)
```
lyrixx, line 47: missing cascade delete

src/Entity/Notifications/Notification.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        return $this->id;
+    }
+
+    public function getTargetUser(): User
+    {
+        return $this->targetUser;
+    }
+
+    public function setTargetUser(User $targetUser): self
```
lyrixx, line 83: I guess we don't need all theses setter. everything can goes to the constructor.

src/Entity/Notifications/ParticipationStatusChangedNotification.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        return $this->participation;
+    }
+
+    public function getEmitter(): User
+    {
+        return $this->emitter;
+    }
+
+    public function setEmitter(User $emitter): self
```
lyrixx, line 42: same

src/EventListener/EasyAdminEventListener.php
```diff
@@ -44,7 +48,8 @@ public function onPostUpdate(GenericEvent $event): void
         $entity = $event->getSubject();
         if ($entity instanceof Submit) {
             if ($entity->hasStatusChanged()) {
-                $this->eventDispatcher->dispatch(new SubmitStatusChangedEvent($entity));
+                $this->eventDispatcher->dispatch(new SlackSubmitStatusChanged($entity));
+                $this->eventDispatcher->dispatch(new NotificationSubmitStatusChangedEvent($entity));
```
lyrixx, line 52: This is wrong :) 

**One event, many listeners**

`SlackSubmitStatusChanged` this name was wrong in the first place. It should have been `SubmitUpdatedEvent`

src/Entity/User.php
```diff
@@ -408,4 +417,29 @@ public function unserialize($serialized): ?array
...
+    {
+        if (!$this->notifications->contains($notification)) {
+            $this->notifications[] = $notification;
+            $notification->setTargetUser($this);
+        }
+
+        return $this;
+    }
+
+    public function removeNotification(Notification $notification): self
```
lyrixx, line 439: Never used?

src/EventListener/EasyAdminEventListener.php
```diff
@@ -31,8 +34,9 @@ public function __construct(
     public static function getSubscribedEvents()
     {
         return [
-            EasyAdminEvents::POST_UPDATE => 'onPostUpdate',
+            EasyAdminEvents::PRE_UPDATE => 'onPostUpdate',
```
lyrixx, line 37: you listen preUpdate, but the method is called onPostUpdate :/

src/EventListener/Notification/NotificationsEventListener.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /** @return array<string> */
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            SubmitAddedEvent::class => 'onSubmitAdded',
+            SubmitStatusChangedEvent::class => 'onSubmitStatusChanged',
+            ParticipationStatusChangedEvent::class => 'onParticipationStatusChanged',
+            NewFeaturedConferenceEvent::class => 'onNewFeaturedConference',
```
lyrixx, line 36: Lack of consistency:

SubmitAdded vs NewFeaturedConference (add vs new)

src/EventListener/Notification/NotificationsEventListener.php
```diff
@@ -0,0 +1,84 @@
...
+        ];
+    }
+
+    public function onSubmitAdded(SubmitAddedEvent $event)
+    {
+        foreach ($event->getSubmit()->getUsers() as $targetUser) {
+            if ($targetUser === $this->security->getUser()) {
```
lyrixx, line 43: save the $user variable outside the loop

src/Repository/NotificationRepository.php
```diff
@@ -0,0 +1,48 @@
...
+        return $this->createQueryBuilder('n')
+            ->update()
+            ->set('n.read', ':true')
+            ->andWhere('n.targetUser = :user')
+            ->andWhere('n.read = :false')
+            ->setParameters([
+                'true' => true,
+                'false' => false,
+                'user' => $user,
+            ])
```
lyrixx, line 43:
```suggestion
            ->set('n.read', ':newRead')
            ->andWhere('n.targetUser = :user')
            ->andWhere('n.read = :currentRead')
            ->setParameters([
                'newRead' => true,
                'currentRead' => false,
                'user' => $user,
            ])
```

(namming a variable `true` is not a good idead)

src/Twig/NotificationsFilter.php
```diff
@@ -0,0 +1,30 @@
...
+use Doctrine\Common\Collections\Collection;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+class NotificationsFilter extends AbstractExtension
+{
+    public function getFunctions()
+    {
+        return [
+            new TwigFunction('getUnreadNotifications', [$this, 'getUnread']),
```
lyrixx, line 14: It should have be a filter be a filter, **but** this is a bad idead to do like that.
In 5 years, when we will have 1K notification, we will load 1000 notifications from the DB for nothing.

Instead, you need to use a dedicated controller (with a nice SQL query) and `render(controller(xxx))` where you want to return your content https://symfony.com/doc/current/templates.html#embedding-controllers

templates/_header.html.twig
```diff
@@ -21,13 +21,44 @@
...
+                <ul class="navbar-nav mr-auto position-relative">
                     <li class="nav-item mr-2">
-                        <a class="nav-link btn btn-outline-secondary btn-sm text-light"
+                        <a class="nav-link btn btn-outline-secondary btn-sm text-light fsize-medium"
                            href="{{ app.user ? path('user_account') : path('login') }}">
                             {{ app.user ? '👤 Account' : '🔒 Login' }}
                         </a>
                     </li>
+                    {% if app.user %}
+                        {% set unreadNotifications = getUnreadNotifications(app.user.notifications) %}
```
lyrixx, line 32: This will change (personne note to track future diff)

templates/user/notifications/new_featured_conference.html.twig
```diff
@@ -0,0 +1,2 @@
+New featured conference ! See it here : <a href="{{ path('conferences_show', { slug: notification.conference.slug }) }}" class="text-primary">See Conference</a>.
```
lyrixx, line 1:
```suggestion
New featured conference! See it here : <a href="{{ path('conferences_show', { slug: notification.conference.slug }) }}" class="text-primary">See Conference</a>.
```

I think you can rework all message. There are too "raw".
And in this case it'll render as:

`New featured conference ! See it here : See Conference.`

**Maybe** something like `A new featured conference has been added` (and the link on all the text).

lyrixx (commented):
Not bad 👏🏼 But there are still many things to keep in mind :) (and to fix)

lyrixx:
image

too much red. Red a used for something  bad / wrong.

You can take example on github, twitter, etc if you want some inspiration  (but use green / blue is better)

src/Entity/Conference.php
HedicGuibert: I did not run the CS fixer yet since it was still a WIP, as I didn't run phpstan or tests yet aswell.
Since there will still be plenty of modifications, I intend to do it once things are more stable.

However, if you feel this is making it uncomfortable to read, I can run at least the fixer

src/Entity/Notifications/Notification.php
HedicGuibert: Since we also have an `emitter` on some child entities who is a user as well, it's more clear (in my opinion) to have a bit more precisions about who does what

src/EventListener/EasyAdminEventListener.php
HedicGuibert: Well it is in reality `SubmitStatusChanged`, but since I had two of those in the same EventListener, I had to make aliases :).

But sure the listeners are quite messy, we could clean this up.

lyrixx: The alias is really a bad idea :) the intent of listener should not leak in the event.
The event is generic

assets/js/urlBuilder.js
```diff
@@ -0,0 +1,13 @@
+const allUrlToBuild = document.getElementsByClassName('js-build-url');
```
lyrixx, line 1: This file has a name in camel case, and assets/js/highlight_scroll.js is in snake case.

you need to choose a camp (snake or kebab)

---

I would name the var `urls`

lyrixx, line 1: This file has a name in camel case, and assets/js/highlight_scroll.js is in snake case.

you need to choose a camp (snake or kebab)

---

I would name the var `urls`

assets/js/urlBuilder.js
```diff
@@ -0,0 +1,13 @@
...
+
+if (allUrlToBuild.length) {
+    const currentUrl = new URL(window.location.href);
+
+    [...allUrlToBuild].forEach(form => {
+        const targetUrl = form.dataset.action;
```
lyrixx, line 7: why do you use data-action, and not directly the action? 
It would be more more accessible I guess

src/EventListener/Notification/NotificationsEventListener.php
```diff
@@ -30,55 +36,67 @@ public function __construct(
...
 
-            $notification = new SubmitAddedNotification($event->getSubmit(), $targetUser);
-            $notification->setEmitter($targetUser);
-            $notification->setTrigger(Notification::TRIGGER_SUBMIT_ADDED);
+            new NewSubmitNotification(
+                submit: $event->getSubmit(),
+                emitter: $currentUser,
+                targetUser: $targetUser,
+                trigger: AbstractNotification::TRIGGER_NEW_SUBMIT
+            );
```
lyrixx, line 60: What do you do with this var?

I have a mixed feeling about this. You relly on the the cascade: persist on the user.notification to make it work.
I would force a persist here to be explicit.

And IIRC, you move all "flush" before dispatching in controllers, so this is never flush?

templates/base.html.twig
```diff
@@ -19,6 +19,7 @@
         {% endblock %}
         {% block javascripts %}
             {{ encore_entry_script_tags('js/app') }}
+            {{ encore_entry_script_tags('js/urlBuilder') }}
```
lyrixx, line 22: Let's include the file urlBuilder in the app file to avoir aving many encore entries since it's always required

lyrixx (commented):
Much much better :clap: :clap:

src/EventListener/Notification/NotificationsEventListener.php
HedicGuibert: Oh you're right, I originally didn't want to dispatch an event if the flush somehow didn't work, that's why I moved it here but this is true that this is now never flushed.

Is it bad to rely on the cascade if there is one ?

lyrixx: It's really personal, but I don't like it. From time to time, bad thing happens.

And anyway:

> explicit is better than implicit

rule 2 of "zen of python" https://www.python.org/dev/peps/pep-0020/ (**MUST READ** / must learn / must apply !)

assets/js/urlBuilder.js
HedicGuibert: It's because everyone knows when they see an element with a dataset that this element will be manipulated with Javascript, which is not clear if I directly use the `action` attribute.

You could argue that this is already pretty clear that this will be manipulated by Javascript thanks to the `js-build-url` class, and you'd be right.

And I could also be wrong on the reason I gave, I don't know

HedicGuibert:
I noticed the `cancelled_participation` route\page doesn't exist, and for this reason I couldn't dispatch an event when someone cancels a participation for some other user.

So I'll have to add this page and dispatch an event when this happens.

assets/js/urlBuilder.js
lyrixx: If people don't have js, if you use action only, it degrades really smooothly

src/Controller/UserAccount/ParticipationController.php
```diff
@@ -73,10 +73,10 @@ public function userParticipation(Request $request): Response
         ]);
     }
 
-    #[Route(path: '/user/future-participations', name: 'future_participations')]
+    #[Route(path: '/user/future-participations', name: 'accepted_participations')]
```
lyrixx, line 76: This move is weird :/ 

There is an inconsistency between the 
* path & methd name
* VS name & template name

webpack.config.js
```diff
@@ -14,6 +14,8 @@ Encore
     .enableSourceMaps(!Encore.isProduction())
     .enableVersioning(Encore.isProduction())
     .addEntry('js/app', './assets/js/app.js')
+    .addEntry('js/highlightScroll', './assets/js/highlightScroll.js')
+    .addEntry('js/urlBuilder', './assets/js/urlBuilder.js')
```
lyrixx, line 18: This should be remvoed, right?

src/Controller/UserAccount/ParticipationController.php
HedicGuibert: It seems I forgot to change the path name, however the template name was already `accepted_participation`.
I made this change for consistency because we are naming things after our workflow constants pretty much everywhere but here.

It was also error prone as well as a bit misleading so I'll just double check I didn't forget any other `futre_xxx` stuff

HedicGuibert:
I ended up noit doing what I said here : https://github.com/jolicode/starfleet/pull/216#discussion_r705104543.

I just reverted my change and made sure we handle the case of different participations marking better. In the end it was the new name which was misleading.

Anyway everything is ready, so I'll let you review and I'll squash if you approve

webpack.config.js
```diff
@@ -14,6 +14,7 @@ Encore
     .enableSourceMaps(!Encore.isProduction())
     .enableVersioning(Encore.isProduction())
     .addEntry('js/app', './assets/js/app.js')
+    .addEntry('js/highlightScroll', './assets/js/highlightScroll.js')
```
lyrixx, line 17: Should be removed, right?

lyrixx:
Thanks for the hard work 💪🏼

PR #219, opened 2021-09-09
Fix multiple display/UX small issues

HedicGuibert:
This adresses several small issues I noticed while working on some user account features.

List of what it fixes/improves :

 * Improve Header buttons
new_buttons

 * Improve dates text
new_date

 * Adds links to user account and front in the admin
new_links

 * Sort talks better in selects, use optgroup to put talks the current user already gave at the top of the list
 * Improve links underline display
new_link_display

 * Fix display issue due to `position: absolute` for flash messages
 * Sort Participations better in participation page (sort them by `id` rather than `createdAt` which wasn't working properly)

### DIFF
--- a/assets/css/app.scss
+++ b/assets/css/app.scss
@@ -55,6 +55,24 @@ header nav {
     }
 }
 
+.header-button {
+    display: block;
+    padding: 0.7em;
+    line-height: 1.5;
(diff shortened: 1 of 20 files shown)


lyrixx (approved):

lyrixx:
Thanks

PR #226, opened 2021-09-28
Fix Google and GitHub login issues

HedicGuibert:
As stated in the title, fixes the login issues we had with GitHub and Google

### DIFF
--- a/src/Security/Authenticator/GitHubAuthenticator.php
+++ b/src/Security/Authenticator/GitHubAuthenticator.php
@@ -12,6 +12,7 @@
 namespace App\Security\Authenticator;
 
 use App\Entity\User;
+use App\Repository\UserRepository;
 use Doctrine\ORM\EntityManagerInterface;
 use KnpU\OAuth2ClientBundle\Client\ClientRegistry;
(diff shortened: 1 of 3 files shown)


src/Security/Authenticator/GitHubAuthenticator.php
```diff
@@ -71,27 +72,38 @@ public function authenticate(Request $request): PassportInterface
...
                 }
 
-                /** @var User|null $user */
-                $user = $this->em->getRepository(User::class)
-                    ->findOneBy(['email' => $githubUser->getEmail()]);
+                $response = $this->httpClient->request('GET', 'https://api.github.com/user/emails', [
+                    'headers' => [
+                        'Authorization' => sprintf('token %s', $accessToken->getToken()),
+                    ],
+                ]);
```
welcoMattic, line 79: The email is no more in the received payload from GitHub? In the `GithubResourceOwner` object?

HedicGuibert: If I understood the API correctly, the `GithubResourceOwner` only contains public emails, but not everyone has its email as public. With my account at least, the `email` was set to `null`

lyrixx (approved):
👍🏼

PR #229, opened 2021-09-29
Add messages in user account when empty data

HedicGuibert:
This adds messages in the user account for cases where the user lacks data.

### DIFF
--- a/templates/user/future_conferences/future_conferences.html.twig
+++ b/templates/user/future_conferences/future_conferences.html.twig
@@ -48,6 +48,8 @@
                         </div>
                     </div>
                 </div>
+            {% else %}
+                <em>No Featured Conference for now.</em>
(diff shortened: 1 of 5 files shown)


templates/user/future_conferences/future_conferences.html.twig
```diff
@@ -11,6 +11,9 @@
 
         <h3 class="ml-5 my-5 text-secondary indent-50">Featured Conferences</h3>
         <div class="d-flex flex-row flex-wrap justify-content-center" id="featured-conferences-block">
+            {% if not featuredConferences|length %}
+                <em>No Featured Conference for now.</em>
+            {% endif %}
             {% for conference in featuredConferences %}
```
lyrixx, line 17: you can use for else instead

PR #233, opened 2021-09-29
Fix some User Account issues

### LINKED ISSUE #232: Fix preprod issues
Due to our recent preprod deployment, we have some issues to fix (probably more to come) :

- [ ] User menus don't match between the Homepage and the other pages 
- [ ] Edit profile form allows users to change their GitHub or Google ID
- [ ] Edit profile form is not working.

HedicGuibert:
fix #232 

- Fixes the edit profile form
- new homepage matching the rest of the user account better :

homepage

### DIFF
--- a/src/Form/UserAccount/UserProfileType.php
+++ b/src/Form/UserAccount/UserProfileType.php
@@ -21,7 +21,6 @@ use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Form\FormError;
 use Symfony\Component\Form\FormEvent;
 use Symfony\Component\Form\FormEvents;
-use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
 
 class UserProfileType extends AbstractType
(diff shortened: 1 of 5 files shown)


HedicGuibert:
I noticed that I had forgotten few other things as well (one wrong route, and the PR was removing the logout button), so I fixed that aswell.
This should be good now.

src/Form/UserAccount/UserProfileType.php
```diff
@@ -101,7 +93,7 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
         ;
     }
 
-    /** @return FormInterface|void */
+    /** @return FormInterface|null */
     public function onPasswordSubmit(FormEvent $event): mixed
```
lyrixx, line 97: Let's use a real type hint instead of this PHPDoc :/ 

BTW, what's the point to return the form? the return value is never used. let's never return something

templates/user/account.html.twig
```diff
@@ -1,43 +1,9 @@
...
-                    {{ '➡️ Logout' }}
-                </a>
-            </li>
-        </ul>
-    </aside>
-
-    <section class="col-8">
+    {% set section = 'homepage' %}
+    <section class="col-8 mx-auto">
+        {{ parent() }}
```
lyrixx, line 6: Much better 👍🏼

src/Form/UserAccount/UserProfileType.php
HedicGuibert: Actually the value is used by the validator. I can't exactly remember where and how, it would require digging, but I'm 100% sure that if I remove the returns and the user submits an invalid form, it breaks.

PR #234, opened 2021-09-29
Improve tests performance

HedicGuibert:
Tests are currently really slow due to password hashing and database resets, this will speed them up a bit.


HedicGuibert:
Looks good to me, but we will need to remove the `encoder` keys when we merge #237.

This is quite faster now !

Before :
before

Now :
after

lyrixx (approved):

HedicGuibert:
I rewrote a good part of this PR since I had forgotten some `resetDatabase` in several places, improving performances further.

Note to my future self and anyone reading this : never use the `resetDatabase` trait :)

PR #236, opened 2021-09-29
Fix missing event trigger on Talks page in user account

HedicGuibert:
When creating a talk on the user account talks page, if we added multiple users to the submit created it would not trigger an event and therefore a notification for other users.
it should have the same behaviour as when users create a submit with multiple other users.


lyrixx:
Thanks

PR #238, opened 2021-10-01
Fix various blocking issues

HedicGuibert:
There was some leftover bugs on the preprod that we needed to fix before production deployment.

Quick list :
     Admin 
- Remove the possibility to create a new past conference, which was not configured and breaking. Remove possibility to create a new Excluded Conference as well
- Admins could remove the User role of anyone, which would completely break their account.

User Account
- Fix issue when trying to submit or ask participation to a conference with multiple editions in the database. If you tried to submit a talk to the `foo` Conference, if a past `foo` Conference exists it will try to submit to this one and throw an error.
- Marking all notifications as read was not working
- We were missing form error for the user field in the participation form

I also reworded a few things here and there.

### DIFF
--- a/templates/user/submit/submit.html.twig
+++ b/templates/user/submit/submit.html.twig
@@ -19,8 +19,10 @@
 {% endblock %}
 
 {% block _submit_users_label %}
-    <label for="{{ full_name }}" class="d-block">Users</label>
-    {{ form_errors(form) }}
+    <label for="{{ full_name }}" class="d-block">
+        Users
+        {{ form_errors(form) }}
+    </label>
 {% endblock %}
 
 {% block _submit_users_widget %}
--- a/templates/user/submit/submit_form.html.twig
+++ b/templates/user/submit/submit_form.html.twig
@@ -19,7 +19,10 @@
 {% endblock %}
 
 {% block _submit_users_label %}
-    <label for="{{ full_name }}" class="d-block">Users</label>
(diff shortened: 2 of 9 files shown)


templates/user/submit/submit.html.twig
```diff
@@ -19,8 +19,10 @@
 {% endblock %}
 
 {% block _submit_users_label %}
-    <label for="{{ full_name }}" class="d-block">Users</label>
-    {{ form_errors(form) }}
+    <label for="{{ full_name }}" class="d-block">
+        Users
+        {{ form_errors(form) }}
+    </label>
```
welcoMattic, line 25: It's weird to display errors inside a `label` tag

templates/user/submit/submit_form.html.twig
```diff
@@ -19,7 +19,10 @@
 {% endblock %}
 
 {% block _submit_users_label %}
-    <label for="{{ full_name }}" class="d-block">Users</label>
+    <label for="{{ full_name }}" class="d-block">
+        Users
+        {{ form_errors(form) }}
+    </label>
```
welcoMattic, line 25: idem

templates/user/submit/submit.html.twig
HedicGuibert: This is how the bootstrap layout naturally does it, it fits better ike this, or I would have to add some CSS. I found it quite weird too

PR #239, opened 2021-10-01
Add try/catch to prevent 500 if Slack is not responding

HedicGuibert:
Slack is currently having DNS issues, and this is quite embarassing for us since we send it a good amount of requests.

This is not preventing us from doing what we want, but it is never good to have a 500 when you try to set a submit as accepted

### DIFF
--- a/src/Notifiers/Slack/SlackNotifier.php
+++ b/src/Notifiers/Slack/SlackNotifier.php
@@ -18,6 +18,7 @@ use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\RouterInterface;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 class SlackNotifier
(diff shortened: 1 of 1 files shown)


src/Notifiers/Slack/SlackNotifier.php
```diff
@@ -239,9 +239,13 @@ private function buildCfpsBlocks(array $endingCfps): array
...
-        $this->httpClient->request('POST', $this->webHookUrl, [
-            'headers' => ['Content-type' => 'application/json'],
-            'body' => json_encode($blocks),
-        ]);
+        try {
+            $this->httpClient->request('POST', $this->webHookUrl, [
+                'headers' => ['Content-type' => 'application/json'],
+                'body' => json_encode($blocks),
+            ]);
+        } catch (\Exception $exception) {
```
lyrixx, line 247: **Never** use \Exception (nor \Throwable). Everything because harder to debug with that.

lyrixx (commented):
I'm not sure about this...
We silence a bug, so the slack notification will never be fired.
A queue system would be nice, but there is not on this project.

May be it's time to introduce queues ? (messenger + doctrine is quite enough) 

Otherwise, I would let the exception pop.

BTW, slack is almost never down, and ATM we can fix it manually to make it work

src/Notifiers/Slack/SlackNotifier.php
```diff
@@ -239,9 +239,13 @@ private function buildCfpsBlocks(array $endingCfps): array
...
+            $this->logger->log('critical', sprintf('An error occured while trying to request Slack : %s', $exception));
```
lyrixx, line 248: 1. use the critical method, it's done for that
2. never cast an exception to a string in a exception message. the context is here for that
3. critical is too high : https://github.com/Seldaek/monolog/blob/main/doc/01-usage.md#log-levels
```suggestion
            $this->logger->error('An error occured while trying to request Slack, ['context' => $exception]);
```

HedicGuibert:
Okay sure, so maybe we want to close this and open an issue ?

Because if we want to add notifications for email or other applications as well, we will want to make sure they are not throwing us errors, right ?

lyrixx:
This is a bit complicated. It's more a political than a technical decision.

Starfleet is build around slack. So I see slack as a first citizen of the startfleet stack.
So IMHO, if slack is down, it's legit to have starfeet down.

And as state, we can fix this issue very simply with one command on the preprod / production server.

So IMHO, my prefered choices (ordered)

1. Use a queue to have native retry (But I think you won't be able to achieve this before the end of the day)
2. Patch the server and close this PR
3. Merge this PR

=> Could you do the second point, and close this PR?

In another application, I would have merged your PR.

Issue #240, opened 2021-10-01
Past submit with accepted status are listed on future submits on user account


welcoMattic:
Past submit with accepted status are listed on future submits on user account.
They shouldn't be there.

PR #241, opened 2021-10-01
~Fixed past sumbit listed in future~ Reverted, it only brings CS stuff

welcoMattic:
Ref https://github.com/jolicode/starfleet/issues/240


lyrixx:
Can you check the tests ?

welcoMattic:
After reconsidering the fix, it's not the right way to do. Submits must be in the right status all the times, it's during the fetch that we have to check if an accepted submit is in the future or not.

lyrixx:
Nothing really change here, right?

welcoMattic:
No, nothing, I choose to revert the fix (which is wrong)

PR #257, opened 2022-02-09
Fix wrong CSS and misplaced no submits block

### LINKED ISSUE #256: Submits go under the footer on user account
On my account, after accepting / rejecting some talks, this happened : 
image

As you can see, the whole block went under the footer.

HedicGuibert:
fix #256 

The else block we wanted to show if there was no submits/participations was at the wrong place and the CSS was causing issues for almost no benefits so I just removed it.

### DIFF
--- a/templates/user/participation/participation.html.twig
+++ b/templates/user/participation/participation.html.twig
@@ -19,17 +19,6 @@
 
 {% block content %}
     <aside class="col-6">
-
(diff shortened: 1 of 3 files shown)


lyrixx (approved):

lyrixx:
Tu peux auto merge + deploy

